PR #1, opened 2020-03-18
Migration of panther_analysis_tool to dedicated repo

nhakmiller:
Moving the `panther_analysis_tool` from the `panther-analysis` repo to here so it can have a dedicated space.


README.md
```diff
@@ -1,2 +1,285 @@
...
+  <i>Command Line Panther Analysis Management</i>
+</p>
+
+<p align="center">
+  <a href="https://docs.runpanther.io">Documentation</a> |
+  <a href="https://docs.runpanther.io/quick-start">Quick Start</a>
+</p>
+
+<p align="center">
+  <a href="https://gitter.im/runpanther/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge"><img src="https://badges.gitter.im/runpanther/community.svg" alt="Gitter"/></a>
```
austinbyers: Remove Gitter link, replace with Slack

README.md
```diff
@@ -1,2 +1,285 @@
...
+  <a href="https://circleci.com/gh/panther-labs/panther-analysis"><img src="https://circleci.com/gh/panther-labs/panther-analysis.svg?style=svg" alt="CircleCI"/></a>
```
austinbyers: Change to point to `panther_analysis_tool` (both the link and the img src)

README.md
```diff
@@ -1,2 +1,285 @@
...
+  <a href="https://cla-assistant.io/panther-labs/panther-analysis" alt="CLA Assistant"><img src="https://cla-assistant.io/readme/badge/panther-labs/panther-analysis"/></a>
```
austinbyers: I'm not sure if we have CLA enabled for this repo yet, but if so change the link

buildspec.yml
```diff
@@ -0,0 +1,11 @@
+version: 0.2
```
austinbyers: We already have a `.circlei/config.yml` file, I don't think we need this one?

config.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
```
austinbyers: Again, don't think we need this file

README.md
nhakmiller: I don't think so either, removing it for now.

PR #6, opened 2020-03-19
Support Reports Field

nhakmiller:
### Background

We're going to be adding support soon for a reports field, which will help report generation. Since we want to move some information currently encoded in tags to the reports field, this change just allows the field to be present and not throw an error during testing. The Panther backend throws away the Reports field when processing the configuration.

### Changes

* Allow a reports field to exists, which will be a map of strings (report names, which will also later be treated as tags) to objects (typically either a list of strings or a dictionary)

### Testing

* Added some different configurations of report fields to some policies I'm working on locally and tested and uploaded them, no issues

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -80,7 +80,9 @@ SPEC_SCHEMA = Schema(
             str,
         Optional('Suppressions'): [str],
         Optional('Tags'): [str],
-        Optional('Reports'): {str: object},
+        Optional('Reports'): {
+            str: object
+        },
         Optional('Tests'): [{
             'Name': str,
             'ResourceType': str,
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -80,7 +80,7 @@ SPEC_SCHEMA = Schema(
             str,
         Optional('Suppressions'): [str],
         Optional('Tags'): [str],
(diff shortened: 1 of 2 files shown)


panther_analysis_tool/main.py
```diff
@@ -80,12 +80,15 @@ def get(self, arg: str, default: Any = None) -> Any:
             str,
         Optional('Suppressions'): [str],
         Optional('Tags'): [str],
+        Optional('Reports'): {
```
jacknagz: Can we also fix the PolicyID vs RuleID issues in the schema?

jacknagz: ResourceType vs LogType

nhakmiller: That's really more of a backend Panther change than a change the `panther_analysis_tool` needs, the backend currently doesn't have a concept of RuleID or LogType so we could let the `panther_analysis_tool` support these fields, but anyone who actually used them wouldn't be able to upload those rules to Panther.

kostaspap:
Something I just realized, we need to add support for the `DedupPeriodMinutes` field. Right now it is default to 60 minutes, but people should be able to change it eventually. They can do it from the UI with https://github.com/panther-labs/panther/pull/475, so we should eventually let them configure it from the analysis tool as well!

PR #8, opened 2020-03-23
Update Phrasing to be more General

nhakmiller:
### Background

Updated the `panther_analysis_tool` to refer more generally to both policies and rules, where previously it was more policy centric.

### Changes

* Updated `panther_analysis_tool` to refer to `analysis` or `policies and rules` in place of just `policies` both in command output and also internally
* Updated docs accordingly

### Testing

* `make ci`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,4 +31,4 @@ unit:
 	pipenv run nosetests -v
 
 integration:
-	panther_analysis_tool test --policies tests/fixtures/valid_policies/
+	panther_analysis_tool test --analysis tests/fixtures/valid_policies/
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -31,4 +31,4 @@ unit:
 	pipenv run nosetests -v
 
 integration:
-	panther_analysis_tool test --policies tests/fixtures/valid_policies/
+	panther_analysis_tool test --analysis tests/fixtures/valid_policies/
```
jacknagz: how about `path`?

PR #19, opened 2020-04-22
Added support for globals

nhakmiller:
### Background

Added support for the new `global` analysis type. Internal re-factoring to give better assuredness of correctness. Drop support for incorrectly named fields.

### Migration Notes

Do to the change in how global helper functions are implemented on the panther backend, the `panther_analysis_tool` will no longer be supporting the `aws_globals` policy as a means of defining globally accessible functions and variables. Instead, the `global` analysis type is taking on this functionality. How to convert an existing `aws_globals` policy or rule into a `global` analysis type:

1. In the specification file, change the value of `AnalysisType` to `global`
1. In the specification file, remove all fields besides `GlobalID`, `FileName`, `Description` (optional), and `Tags` (optional)
1. While not required, it is recommended you remove the `def policy(_): return True` or `def rule(_): return True` stub function from the global body as it is no longer required.
1. Update all references to `import aws_globals` to `import panther` (and update the corresponding usages)
1. Upload your new rules, policies, and globals

### Changes

* The `test` command now correctly categorizes `globals` as a separate entity and imports them first.
* Schema validation is now more strict, enforcing e.g. `PolicyID` on policies vs. `RuleID` on rules vs. `GlobalID` on globals. Previously a rule could have a `PolicyID` and vice versa.
* Dropped support for `aws_globals`
* Improved error reporting when communicating with Panther directly

### Testing

* `make ci`


panther_analysis_tool/main.py
```diff
@@ -51,32 +49,76 @@ def get(self, arg: str, default: Any = None) -> Any:
...
+        Optional('Description'): str,
+        Optional('Tags'): [str],
+    },
+    ignore_extra_keys=False)
+
+POLICY_SCHEMA = Schema(
     {
         'AnalysisType':
-            Or("policy", "rule"),
+            Or("policy"),
```
nhakmiller, line 69: There may be a cleaner way to do this, open to suggestions.

PR #22, opened 2020-05-01
Add support for generic filtering

### LINKED ISSUE #5: Support Tag Filtering
Migrated from `panther-analysis`.

Rules and Policies both support tagging, the Panther Analysis tool could accept a list of tags that would act as a filter. The format would be something like:

`$ panther-cli test --policies analysis --tags CIS PCI`

The result would be the panther-cli would then run tests on only the policies/rules that were tagged with either CIS or PCI. This would be even more useful with the zip command:

`$ panther-cli zip --policies analysis --tags CIS`

This command would succinctly bundle all policies and rules that are tagged CIS, which would make creating bundles of just desired policies/rules easier without having to do a bunch of file copying to get just the policies/rules that are relevant into the right place.

nhakmiller:
### Background

Adds the `--filter` flag to all commands, allowing users to filter to just specific analysis objects by specifying any valid field in an analysis specification file.

Closes #5 .

Example usage:

`panther_analysis_tool test --path tests/fixtures/valid_policies --filter AnalysisType=policy,global`
  - will only apply to `policy` and `global` analysis types, excluding rules

`panther_analysis_tool test --path tests/fixtures/valid_policies --filter AnalysisType=policy Severity=Critical`
  - will only apply to `policy` analysis types that have the severity `Critical`

`panther_analysis_tool test --path tests/fixtures/valid_policies --filter AnalysisType=policy,global Severity=Critical,""`
  - By providing an empty string, you can indicate that the field need not be present. So this will only apply to `policy` and `global` analysis types, and only if they have a severity of `Critical` or no severity specified at all. In this case this is necessary because `global` analysis types don't have a severity. It is also useful for optional fields.

### Changes

* Added generic filtering to all commands

### Testing

* `make ci`
* manually verifying correctness

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -222,18 +222,15 @@ def zip_analysis(args: argparse.Namespace) -> Tuple[int, str]:
     # The colon character is not valid in filenames.
     current_time = datetime.now().isoformat(timespec='seconds').replace(
         ':', '-')
-    filename = 'panther-analysis'
-    zip_out = zipfile.ZipFile('{}-{}.zip'.format(filename, current_time), 'w',
-                              zipfile.ZIP_DEFLATED)
+    filename = 'panther-analysis-{}.zip'.format(current_time)
+    with zipfile.ZipFile(filename, 'w', zipfile.ZIP_DEFLATED) as zip_out:
+        analysis = filter_analysis(list(load_analysis_specs(args.path)),
+                                   args.filter)
+        for analysis_spec_filename, dir_name, analysis_spec in analysis:
+            zip_out.write(analysis_spec_filename)
+            zip_out.write(os.path.join(dir_name, analysis_spec['Filename']))
 
-    analysis = filter_analysis(list(load_analysis_specs(args.path)),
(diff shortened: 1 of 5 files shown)


panther_analysis_tool/main.py
```diff
@@ -223,9 +223,17 @@ def zip_analysis(args: argparse.Namespace) -> Tuple[int, str]:
     current_time = datetime.now().isoformat(timespec='seconds').replace(
         ':', '-')
     filename = 'panther-analysis'
-    return 0, shutil.make_archive(
-        os.path.join(args.out, '{}-{}'.format(filename, current_time)), 'zip',
-        args.path)
+    zip_out = zipfile.ZipFile('{}-{}.zip'.format(filename, current_time), 'w',
```
austinbyers, line 226: ZipFile is a context manager, so instead of having to `.close()` it explicitly, you can say:

```python
with zipfile.ZipFile(...) as zip_out:
    zip_out.write(...)
```

This will also close the file if there were any exceptions

nhakmiller: Good call!

PR #25, opened 2020-05-16
Show dedup/title values

### LINKED ISSUE #24: Test Dedup and Title Strings
While running tests, we need to validate the dedup() and title() functions. We can do it like this:

1. Failing the test if dedup or title throws an error and explain why -> [ERROR] KeyError in dedup()
2. If not, show the generated title next to the result -> [PASS] User [FooBar] has Logged Into server [instance-1]

nhakmiller:
### Background

When testing rules, show the title and dedup field if one is set. Closes #24 .

Added support for multiple globals, and require the `GlobalID` field in global analysis types now. Also automatically searches for globals in the default location in addition to the path specified.

### Changes

* When testing rules, show the title and dedup values if one is set
* Must specify the `GlobalID` field in global analysis types now, and the `GlobalID` will be used as the module name for importing
* Automatically checks the path `global_helpers` (relative to where the command is being run from) for ANY analysis items. Recommend only putting global analysis types here.

### Testing

* make ci

Example output:

```
$ panther_analysis_tool test --path tests/fixtures/valid_policies
[INFO]: Testing analysis packs in tests/fixtures/valid_policies

AWS.IAM.MFAEnabled

AWS.IAM.BetaTest

AWS.CloudTrail.MFAEnabled
		[Title] root does not have MFA enabled
		[Dedup] root
		[Title] test does not have MFA enabled
		[Dedup] test
```

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -32,6 +32,7 @@ import yaml
 
 import boto3
 
+HELPERS_LOCATION = 'global_helpers'
 
 class TestCase():
 
@@ -58,6 +59,8 @@ GLOBAL_SCHEMA = Schema(
     {
         'AnalysisType': Or("global"),
         'Filename': str,
+        'GlobalID':
+            str,
         Optional('Description'): str,
         Optional('Tags'): [str],
     },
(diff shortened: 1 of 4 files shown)


nhakmiller:
CI is failing because I have 16/15 local variables in a function, but if you want to check out this branch and use locally for the time being until I sort that out it should be working just fine.

PR #26, opened 2020-06-08
added better exception handling & filtering

nhakmiller:
### Background

This change adds better exception handling to the `panther_analysis_tool` so you won't get dumped to a help message when processing bad python/YAML quite so often.

Additionally, this change adds better support for filtering based on tags or any other field that accepts a list of values.

### Changes

* Adds generic exception handling to:
  * Loading & validating specification files
  * Executing policy/rule tests
  * Running the main script body
* Adds better filtering for fields that contain lists
  * Previously if a field contained a list, that filter would have to specify EVERY value in that list to be considered a match. Now only one field needs to be specified

### Testing

* Unit tests
* Manual testing of a few known issues

### Examples

* Filtering based on tags
```
% panther_analysis_tool test --path aws_redshift_policies --filter Tags="Data Protection"
[INFO]: Testing analysis packs in aws_redshift_policies

AWS.Redshift.Cluster.Encryption
	[PASS] Encryption Enabled
	[PASS] Encryption Not Enabled
```

* Filtering based on multiple tags
```
% panther_analysis_tool test --path aws_redshift_policies --filter Tags="Data Protection","Security Control"
[INFO]: Testing analysis packs in aws_redshift_policies

AWS.Redshift.Cluster.Encryption
	[PASS] Encryption Enabled
	[PASS] Encryption Not Enabled

AWS.Redshift.Cluster.Logging
	[PASS] Logging Disabled
	[PASS] Logging Enabled

AWS.Redshift.Cluster.VersionUpgrade
	[PASS] Allow Version Upgrade Not Set
	[PASS] Allow Version Upgrade
	[PASS] Does Not Allow Version Upgrade
```

* policy with an unexpected error
```
% panther_analysis_tool test --path aws_redshift_policies --filter Tags="Data Protection"
[INFO]: Testing analysis packs in aws_redshift_policies

AWS.Redshift.Cluster.Encryption
[WARNING]: Unexpected exception: "division by zero"
[WARNING]: Unexpected exception: "division by zero"
```

* policy with an invalid specification file
```
panther_analysis_tool test --path aws_redshift_policies --filter Tags="Data Protection"
[INFO]: Testing analysis packs in aws_redshift_policies

AWS.Redshift.Cluster.Encryption
	[PASS] Encryption Enabled
	[PASS] Encryption Not Enabled

Invalid: aws_redshift_policies/abc.yml
	';laskdjvnc,mmblakdjshbflaksdfkj' should be instance of 'dict'

[ERROR]: [('aws_redshift_policies/abc.yml', SchemaUnexpectedTypeError("';laskdjvnc,mmblakdjshbflaksdfkj' should be instance of 'dict'"))]
```

* testing a directory where no filters are matched
```
% panther_analysis_tool test --path aws_redshift_policies --filter Tags="AWSB"
[INFO]: Testing analysis packs in aws_redshift_policies

[ERROR]: ["No analyses in aws_redshift_policies matched filters {'Tags': ['AWSB']}"]
```

* testing a directory with no analysis in it
```
% panther_analysis_tool test --path no_policies
[INFO]: Testing analysis packs in no_policies

[ERROR]: ['Nothing to test in no_policies']
```

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -477,7 +477,7 @@ def setup_parser() -> argparse.ArgumentParser:
         prog='panther_analysis_tool')
     parser.add_argument('--version',
                         action='version',
-                        version='panther_analysis_tool 0.3.0')
+                        version='panther_analysis_tool 0.3.1')
     subparsers = parser.add_subparsers()
 
     test_parser = subparsers.add_parser(
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -311,9 +311,15 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
         list(load_analysis_specs(args.path)) +
         list(load_analysis_specs(HELPERS_LOCATION)))
 
+    if len(analysis) == 0:
+        return 1, "No valid analyses found in {}".format(args.path)
+        
     # Apply the filters as needed
     global_analysis = filter_analysis(global_analysis, args.filter)
     analysis = filter_analysis(analysis, args.filter)
+
+    if len(analysis) == 0:
+        return 1, "No analyses in {} matched filters {}".format(args.path, args.filter)
 
     # First import the globals
     for analysis_spec_filename, dir_name, analysis_spec in global_analysis:
@@ -375,9 +381,15 @@ def filter_analysis(analysis: List[Any], filters: Dict[str, List]) -> List[Any]:
 
     filtered_analysis = []
     for file_name, dir_name, analysis_spec in analysis:
-        if all(
-                analysis_spec.get(key, "") in values
-                for key, values in filters.items()):
+        match = True
+        for key, values in filters.items():
+            spec_value = analysis_spec.get(key, "") 
+            spec_value = spec_value if type(spec_value) is list else [spec_value]
+            if not set(spec_value).intersection(values):
+                match = False
+                break
+
+        if match:
             filtered_analysis.append((file_name, dir_name, analysis_spec))
 
     return filtered_analysis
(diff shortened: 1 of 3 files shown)


panther_analysis_tool/main.py
```diff
@@ -311,10 +311,16 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
         list(load_analysis_specs(args.path)) +
         list(load_analysis_specs(HELPERS_LOCATION)))
 
+    if len(analysis) == 0:
+        return 1, "No valid analyses found in {}".format(args.path)
```
jacknagz, line 315:
```suggestion
        return 1, "No tests found in {}, skipping!".format(args.path)
```

jacknagz: what do you think about `tests` vs `analyses`?

panther_analysis_tool/main.py
```diff
@@ -410,6 +422,11 @@ def classify_analysis(
                 SchemaUnexpectedTypeError) as err:
             invalid_specs.append((analysis_spec_filename, err))
             continue
+        except Exception as err:  # pylint: disable=broad-except
+            # Catch arbitrary exceptions thrown by bad specification files
+            logging.warn('Unexpected schema validation error "{}", please report this error message to our public repo'.format(err))
```
jacknagz, line 427:
```suggestion
            logging.warn('Unexpected config validation error "{}"'.format(err))
```

What's the logic around posting to our public repo? What do you mean

nhakmiller: My thought process is that there are some exceptions we expect from time to time (such as bad schemas), and we're ok with discarding those immediately. But there may be some other exceptions that we want to handle in different ways, we just don't know that because we've never run into them before. I can just treat those exceptions the same as the ones we are expecting to happen and ignore them, but then it just might be slightly harder to track down certain issues in the future. I can remove the message.

panther_analysis_tool/main.py
nhakmiller: I changed to something similar to this.

PR #33, opened 2020-07-02
Fix ResourceType/LogType field name in README examples

giorgosp:
Simple bug fix of `ResourceType`/`LogType` field name in README examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ Reference: https://www.link-to-info.io
 Tests:
   -
     Name: Name to describe our first test.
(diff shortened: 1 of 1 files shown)


giorgosp:
It's interesting to know that these fields are not really used by the analysis-tool. 
However, they are needed in the policy/rule test API and since the tool can upload policies/rules to a Panther deployment, we have to keep them. 

In the context of https://github.com/panther-labs/panther/issues/610, where we will remove the fields, we probably need to have a note that they are actually needed until Panther version xxx (probably 1.6.0)

PR #34, opened 2020-07-02
Cleanup unneeded ResourceType/LogType fields in policy/rule tests

giorgosp:
**WIP:** 
We should communicate the Panther version after which the policy/rule tests can be deployed without the extra fields.
I have added a TODO in the README to note it there at least.
Will work on the analysis-api now so that it can handle tests without these fields and after we merge that and know to which Panther version it will be included, we can get back and merge this one.

### Background

The ResourceType (in policy tests) and LogType (in rule tests) fields are not
necessary since they can be inferred from the policy/rule under test.

The tool can parse older policies/rules that include these fields.
However, this is a backwards-incompatible change for the analysis-api, so
the fields must be present in the tests when deploying them to older Panther versions.

See https://github.com/panther-labs/panther/issues/610 for more context.

### Changes

Changed the schema to successfully parse tests that include these fields. The fields are now ignored by the tool.
Also the fields have been removed from almost all the fixtures and two more test fixtures have been added to test that
the tool can successfully parse them.

### Testing

`make unit integration`

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ Reference: https://www.link-to-info.io
 Tests:
   -
     Name: Name to describe our first test.
-    ResourceType: Resource.Type.Here
+    ResourceType: Resource.Type.Here  # Not needed in Panther versions >= 1.6.0
     ExpectedResult: true/false
     Resource:
       Key: Values
@@ -288,7 +288,7 @@ Reference: https://www.link-to-info.io
 Tests:
   -
     Name: Name to describe our first test.
-    LogType: Log.Type.Here
+    LogType: Log.Type.Here  # Not needed in Panther versions >= 1.6.0
     ExpectedResult: true/false
     Resource:
       Key: Values
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -65,12 +65,13 @@ POLICY_SCHEMA = Schema(
         },
         Optional('Tests'): [{
             'Name': str,
-            'ResourceType': str,
+            Optional('ResourceType'):
+                str,  # Not needed anymore, optional for backwards compatibility
             'ExpectedResult': bool,
             'Resource': object,
         }],
     },
-    ignore_extra_keys=False)
+    ignore_extra_keys=False)  # Prevent user typos on optional fields
 
 RULE_SCHEMA = Schema(
     {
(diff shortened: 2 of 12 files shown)


panther_analysis_tool/schemas.py
```diff
@@ -65,12 +65,12 @@
...
         Optional('Tests'): [{
             'Name': str,
-            'ResourceType': str,
             'ExpectedResult': bool,
             'Resource': object,
         }],
     },
-    ignore_extra_keys=False)
+    # For backwards compatibility with older schemas having more fields
+    ignore_extra_keys=True)
```
nhakmiller, line 73: My concern with `ignore_extra_keys=True` is that then any keys could be added, not just the test fields right? Perhaps it would be better to just leave the `ResourceType` and `LogType` fields in the schemas but change them to optional. Then we could still catch extraneous fields that are not just there for backwards compatibility, and also allow the backwards compatible fields to stay in.

giorgosp: Yes, all extra fields would be ignored. I couldn't think of any downside for this generally. Maybe if the `analysis-api` is designed the same way and crashes if it sees extra fields? Will check. Could you think of any other issue with ignoring all extra fields?

Thanks for reviewing!

nhakmiller: The main reason is to protect users against mistakes, if they put in a typo like `Descrpition`, no errors would be raised if we accept additional fields because `Description` is optional in the first place and `Descrpition` would just be ignored as an extraneous field. Or alternatively if they put in `Run book` instead of `Runbook`, or they call `Test` instead of `Tests`, those types of mistakes. Perhaps that is being too coddling to users, but I think it is helpful.

I don't think it would cause any issues on the backend in the `analysis-api` though, in fact in the past we've added fields to the policies/rules & `panther_analysis_tool` first before adding them to the `analysis-api` and it didn't cause any issues.

giorgosp: Makes sense! Reverted to fail on extra keys and made the specific fields optional.

README.md
```diff
@@ -223,7 +223,7 @@ Reference: https://www.link-to-info.io
 Tests:
   -
     Name: Name to describe our first test.
-    ResourceType: Resource.Type.Here
+    ResourceType: Resource.Type.Here  # Not needed after Panther version #TODO (the Panther version whose API doesnt need this field)
```
kostaspap, line 226: Typo: `doesn't` (missing apostrophe)

giorgosp: I think we cannot mark it deprecated because users *must* use it with the current/older Panther versions. Then after a target Panther version can they stop using it.

nhakmiller: Almost certainly this will be version 1.6 if you are also working on the `analysis-api` changes this sprint.

giorgosp: Added version 1.6 in the comment and fixed the ci fail.

PR #35, opened 2020-07-10
Fixed up some unexpected behavior

nhakmiller:
### Background

No longer prints the help message unless the user runs the tool with `--help` or `-h`. 

### Changes

* No longer print the help messages unless specifically requested
* If a rule/policy raises an exception report this as a failed test

### Testing

* Unit testing


jacknagz:
@nhakmiller can we add a unit test that uses `--filter` with tags? That's what was failing for me.

nhakmiller:
> @nhakmiller can we add a unit test that uses `--filter` with tags? That's what was failing for me.

I was never able to replicate the failures you were experiencing, even with the exact commands you run. I can add some unit tests though.

PR #36, opened 2020-07-24
Add support for destination overrides

nhakmiller:
### Background

@s0l0ist added support for destination overrides in Panther including in the BulkUpload API, so we should allow users to define this field when managing policies or rules from the commandline.

### Changes

* Add `OutputIds` as an optional field to policies and rules. This field is a list of strings, where each string is the ID of a destination. In a future iteration, we might allow users to specify a destination name that is then mapped to the corresponding destination ID by either the CLI tool or the backend
* Fixed some linting/typing issues `make ci` found 

### Testing

* Unit tests

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -47,7 +47,8 @@ class TestCase():
     def __init__(self, data: Dict[str, Any]) -> None:
         """
         Args:
-            data (Dict[str, Any]): An AWS Resource representation or Log event to test the policy or rule against respectively.
+            data (Dict[str, Any]): An AWS Resource representation or Log event to test the policy
+            or rule against respectively.
         """
         self._data = data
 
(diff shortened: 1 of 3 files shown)


nhakmiller:
> How does a user get the ID?

Unfortunately it's not very easy right now. The only way to get the ID is to look it up in the destinations dynamo table. I discussed this with @s0l0ist and we have a few ideas for improvements:

- A command for the analysis tool to list destinations & their IDs and possibly even automatically convert names -> IDs for any local analysis packs (would require AWS creds)
- An update to the BulkUpload API that would translate the destination name -> ID

I would prefer the second option, but since we're past the code freeze I figured I would at least make it possible to use right now for power users who are ok with checking in dynamo and we can iterate with improvements as we go. If a customer needs it right away we can implement the first option quickly as needed.

austinbyers:
We could also just show the output ID in the web app. You can already find alert IDs this way, for example. That would probably be the fastest fix of all

nhakmiller:
> We could also just show the output ID in the web app. You can already find alert IDs this way, for example. That would probably be the fastest fix of all

That could work too, I just know we already started the release testing for 1.6 and didn't want to delay it further for a small issue like this. I'll put in an issue in Panther suggesting one of these two changes and maybe we can get it for 1.7.

s0l0ist:
Something like this maybe? 
Screen Shot 2020-07-24 at 12 22 45 PM

PR #43, opened 2020-09-02
Lindseyw test dedup title

### LINKED ISSUE #18: Test Support for Dedup and Title
## Description

Currently, the `panther_analysis_tool` only validates the rule function. To ensure that our deduplication and title generation works properly, we should add checks to make sure `None` is not returned.

lindsey-w:
### Background

Overall test cases were passing even if the dedup/title functions returned None. Closes #18 

### Changes

* Added ability for overall test to fail if the dedup or title function return None
* Update output of manually running to show dedup/title check status

### Testing

* Added passing and failing unit tests. 

Example Output

Original Format:
```
% panther_analysis_tool test --path ./tests/fixtures/valid_analysis
[INFO]: Testing analysis packs in ./tests/fixtures/valid_analysis

...

AWS.CloudTrail.MFAEnabled
```

New Format:
```
% panther_analysis_tool test --path ./tests/fixtures/valid_analysis
[INFO]: Testing analysis packs in ./tests/fixtures/valid_analysis

...

AWS.CloudTrail.MFAEnabled
	[PASS] User MFA enabled passes compliance.
		[PASS] [dedup] test
		[PASS] [title] test does not have MFA enabled
```

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -429,6 +429,7 @@ def run_tests(analysis: Dict[str, Any], analysis_funcs: Dict[str, Any],
             test_result['outcome'] = 'FAIL'
 
         # check dedup and title function return non-None
+        # Only applies to rules which match an incoming event
         if unit_test['ExpectedResult']:
             for func in ['dedup', 'title']:
                 if analysis_funcs.get(func):
@@ -436,7 +437,7 @@ def run_tests(analysis: Dict[str, Any], analysis_funcs: Dict[str, Any],
                         test_result[func] = 'FAIL'
                         test_result['outcome'] = 'FAIL'
 
-        if test_result['expected'] == 'FAIL':
+        if test_result['outcome'] == 'FAIL':
             failed_tests[analysis.get('PolicyID') or
                          analysis['RuleID']].append(unit_test['Name'])
 
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -420,10 +420,13 @@ def run_tests(analysis: Dict[str, Any], analysis_funcs: Dict[str, Any],
                          analysis['RuleID']].append(unit_test['Name'])
             continue
 
+        # using a dictionary to map between the tests and their outcomes
+        # assume the test passes (default "PASS") 
+        # until failure condition is found (set to "FAIL")
         test_result = defaultdict(lambda: 'PASS')
         # check expected result
         if result != unit_test['ExpectedResult']:
-            test_result['expected'] = 'FAIL'
+            test_result['outcome'] = 'FAIL'
 
         # check dedup and title function return non-None
         if unit_test['ExpectedResult']:
@@ -431,14 +434,14 @@ def run_tests(analysis: Dict[str, Any], analysis_funcs: Dict[str, Any],
                 if analysis_funcs.get(func):
                     if not analysis_funcs[func](test_case):
                         test_result[func] = 'FAIL'
-                        test_result['expected'] = 'FAIL'
+                        test_result['outcome'] = 'FAIL'
 
         if test_result['expected'] == 'FAIL':
             failed_tests[analysis.get('PolicyID') or
                          analysis['RuleID']].append(unit_test['Name'])
 
         # print results
-        print('\t[{}] {}'.format(test_result['expected'], unit_test['Name']))
+        print('\t[{}] {}'.format(test_result['outcome'], unit_test['Name']))
         for func in ['dedup', 'title']:
             if analysis_funcs.get(func) and unit_test['ExpectedResult']:
                 print('\t\t[{}] [{}] {}'.format(
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -419,16 +419,30 @@ def run_tests(analysis: Dict[str, Any], analysis_funcs: Dict[str, Any],
             failed_tests[analysis.get('PolicyID') or
                          analysis['RuleID']].append(unit_test['Name'])
             continue
-        test_result = 'PASS'
+
+        test_result = defaultdict(lambda: 'PASS')
+        # check expected result
         if result != unit_test['ExpectedResult']:
-            test_result = 'FAIL'
+            test_result['expected'] = 'FAIL'
+
+        # check dedup and title function return non-None
+        if unit_test['ExpectedResult']:
+            for func in ['dedup', 'title']:
+                if analysis_funcs.get(func):
+                    if not analysis_funcs[func](test_case):
+                        test_result[func] = 'FAIL'
+                        test_result['expected'] = 'FAIL'
+
+        if test_result['expected'] == 'FAIL':
             failed_tests[analysis.get('PolicyID') or
                          analysis['RuleID']].append(unit_test['Name'])
-        print('\t[{}] {}'.format(test_result, unit_test['Name']))
-        if analysis_funcs.get('title') and unit_test['ExpectedResult']:
-            print('\t\t[Title] {}'.format(analysis_funcs['title'](test_case)))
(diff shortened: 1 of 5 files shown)


panther_analysis_tool/main.py
```diff
@@ -419,16 +419,30 @@ def run_tests(analysis: Dict[str, Any], analysis_funcs: Dict[str, Any],
...
-        test_result = 'PASS'
+
+        test_result = defaultdict(lambda: 'PASS')
+        # check expected result
         if result != unit_test['ExpectedResult']:
-            test_result = 'FAIL'
+            test_result['expected'] = 'FAIL'
+
+        # check dedup and title function return non-None
+        if unit_test['ExpectedResult']:
```
austinbyers, line 429: What's the purpose of this conditional? Could the expected result be None for a unit test? I wasn't aware of that.

Or do we only care about the dedup/title when the test case is supposed to return True? Keep in mind rules and policies have flipped logic, so a normal/compliant policy is True, but a normal/unmatched rule is False.

I suspect we could remove this line, but I'm not 100% sure

panther_analysis_tool/main.py
```diff
@@ -419,16 +419,30 @@ def run_tests(analysis: Dict[str, Any], analysis_funcs: Dict[str, Any],
             failed_tests[analysis.get('PolicyID') or
                          analysis['RuleID']].append(unit_test['Name'])
             continue
-        test_result = 'PASS'
+
+        test_result = defaultdict(lambda: 'PASS')
+        # check expected result
         if result != unit_test['ExpectedResult']:
-            test_result = 'FAIL'
+            test_result['expected'] = 'FAIL'
```
austinbyers, line 426: This is the actual test result, right? Let's reserve the word 'expected' to mean the expected unit test outcome, not the pass/fail result of running the test. Maybe `test_result['status']` or `test_result['outcome']`, etc.

Perhaps a comment on line 423 about why `test_result` is now a dictionary - what is it mapping between?

tests/fixtures/example_rule_missing_field.yml
```diff
@@ -0,0 +1,34 @@
...
+    ExpectedResult: true
+    Log:
+      {
+        "Arn": "arn:aws:iam::123456789012:user/test",
+        "CreateDate": "2019-01-01T00:00:00",
+        "CredentialReport": {
+          "MfaActive": true,
+          "PasswordEnabled": true
+        }
+      }
```
austinbyers, line 34: You certainly can use JSON syntax here, but you don't have to. This would work as well:

```yml
    Log:
      Arn: arn:aws:iam::123456789012:user/test,
      CreateDate: '2019-01-01T00:00:00',
      CredentialReport:
        MfaActive: true,
        PasswordEnabled: true
```

Nitpick: be sure to add a newline at the end of the file as well (that will make the red dash in GitHub go away)

panther_analysis_tool/main.py
nhakmiller: We only care about the dedup/title when the test case is supposed to be True. You are correct about rules & policies having flipped logic, but this is one of those weird peculiarities between rules & policies because rules are the only ones that support title/dedup functions. Policies don't support these, and the caller function of this function handles that by not passing in a title/dedup function in `analysis_funcs` for policies.

panther_analysis_tool/main.py
nhakmiller: It appears she's using the dictionary to also store whether or not the dedup/title functions are passing, which seems fine to me.

I agree though, I think a comment above line 423 explaining why a default dict with a default value of `PASS` is being used here, and changing `expected` to something like `outcome` makes sense.

nhakmiller:
> ### Testing
> * Added passing and failing unit tests.

When I make changes to the panther_analysis_tool that affect the output, I like to add a little snippet showing how it changed. Something like:

Former output:
```
$ panther_analysis_tool test
ABC
XYZ
```

New output:
```
$ panther_analysis_tool test
ABC
123
```

If you could update the PR description with that it would be helpful I think.

nhakmiller:
Also, you can add closes #18 and GitHub will automatically associate this PR to the issue it closes.

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -62,6 +62,13 @@ def test_with_filters(self):
         assert_equal(return_code, 0)
         assert_equal(len(invalid_specs), 0)
 
+    def test_invalid_rule_definition(self):
+        args = pat.setup_parser().parse_args('test --path tests/fixtures --filter Severity=Critical RuleID=AWS.CloudTrail.MFAEnabled'.split())
+        args.filter = pat.parse_filter(args.filter)
+        return_code, invalid_specs = pat.test_analysis(args)
+        assert_equal(return_code, 1)
+        assert_equal(len(invalid_specs), 1)
```
nhakmiller, line 70: I think we want  `len(invalid_specs)` to be 0 here. `invalid_specs` should only be populated if there are any rules/policies that have invalid specification files.  In this case, I think the specification file should be valid but the tests should fail, which should have a `return_code` of 1 and no `invalid_specs`.

nhakmiller (commented):
Looks pretty good to start, one comment from me about the unit test and then see Austin's comment about naming and it should be good to go!

panther_analysis_tool/main.py
lindsey-w: I made these changes, let me know if you were expecting something different.

tests/unit/panther_analysis_tool/test_main.py
lindsey-w: I tried to create a filter (Severity=Critical RuleID=AWS.CloudTrail.MFAEnabled) for that test that would only pick up the new rule files, but it would continue to show the invalid spec for the file in the same directory (example_malformed_policy.yml). I can remove that check altogether since that particular check is done in a different test case, or is there a different way to filter to just pick up those specific, new files?

panther_analysis_tool/main.py
austinbyers: Ah, maybe a comment to that effect then:

```go
# Only applies to rules which match an incoming event
```

tests/unit/panther_analysis_tool/test_main.py
nhakmiller: Ideally we would just filter to the new files. You've already created a brand new rule body & specification file, so you can just update the `RuleID` for your new rule to be something like `AWS.CloudTrail.MFAEnabled.ReturnsNone`, then use that as your filter. That should filter out to just the test cases you care about.

lindsey-w: I updated the filter, but it was still returning an invalid spec from the directory:

```
% panther_analysis_tool test --path tests/fixtures --filter RuleID=AWS.CloudTrail.MFAEnabled.ReturnsNone
[INFO]: Testing analysis packs in tests/fixtures

AWS.CloudTrail.MFAEnabled.ReturnsNone
        [FAIL] User MFA enabled passes compliance but fails dedup check.
                [FAIL] [dedup] None
                [PASS] [title] None does not have MFA enabled

--------------------------
Panther CLI Test Summary
        Path: tests/fixtures
        Passed: -1
        Failed: 1
        Invalid: 1
```

It looks like the invalid spec check (classify_analysis, line 246) is done prior to the filtering (filter_analysis, line 254), so any invalid spec in the directory will still be flagged in the output? Since this is the failure case, I didn't want to add it to the "valid_analysis" dir.

nhakmiller: Got it. This should be fine then, thanks for looking into it.

nhakmiller (approved):

panther_analysis_tool/main.py
lindsey-w: Added

Issue #37, opened 2020-08-01
Display Name validation is not sufficient


nhakmiller:
Panther does not accept certain special characters in display names, such as:

- `'`
- `"`
- `&`
- `,`

and perhaps others. The panther_analysis_tool should warn against these characters when they are present in a displayname.

3nvi:
Yup, it's all chars that are URL-unsafe 😄

PR #45, opened 2020-09-17
#37 - Display Name Validation

wey-chiang:
Added imports And, Regex from schemas. Added character validation for DisplayName, GlobalID, PolicyID, and RuleID

### Background

Panther does not accept certain special characters in display names, such as:

'
"
&
,
and perhaps others. The panther_analysis_tool should warn against these characters when they are present in a displayname.

### Changes

All changes, with the exception of related testing, that have been made are specific to the `panther_analysis_tool/schemas.py` file.
Created a new global variable `NAME_ID_VALIDATION_REGEX` which is meant to check against strings to ensure the string only contains alphanumeric characters along with dashes, underscores, periods, and spaces.
For GLOBAL_SCHEMA, POLICY_SCHEMA, and RULE_SCHEMA, the schema definition for the fields DisplayName, GlobalID, PolicyID, and RuleID have been updated to verify that the field value is both: of type str and matching NAME_ID_VALIDATION_REGEX

### Testing

Created a copy of tests/fixtures/valid_analysis/policies/example_policy.yml and the respective .py file
Edited the copy to include invalid characters
image

### DIFF
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -16,7 +16,7 @@ limitations under the License.
 
 from schema import And, Optional, Or, Regex, Schema
 
-NAME_ID_VALIDATION_REGEX = Regex(r"^[A-Za-z0-9\-\_\.\ \(\)]+$")
+NAME_ID_VALIDATION_REGEX = Regex(r"^[A-Za-z0-9_. ()-]+$")
 
 TYPE_SCHEMA = Schema({
     'AnalysisType': Or("policy", "rule", "global"),
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -94,7 +94,7 @@ RULE_SCHEMA = Schema(
         Optional('DedupPeriodMinutes'):
             int,
         Optional('DisplayName'):
-            And(str, Regex("")),
+            And(str, NAME_ID_VALIDATION_REGEX),
         Optional('OutputIds'): [str],
         Optional('Reference'):
             str,
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -14,7 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 '''
 
-from schema import Optional, Or, Schema
+from schema import And, Optional, Or, Regex, Schema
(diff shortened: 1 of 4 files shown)


panther_analysis_tool/schemas.py
```diff
@@ -14,7 +14,9 @@
 limitations under the License.
 '''
 
-from schema import Optional, Or, Schema
+from schema import And, Optional, Or, Regex, Schema
+
+NAME_ID_VALIDATION_REGEX = Regex(r"^[A-Za-z0-9\-\_\.\ \(\)]+$")
```
austinbyers, line 19: Nit: inside a character group you typically don't need escaping. This should be equivalent:

```suggestion
NAME_ID_VALIDATION_REGEX = Regex(r"^[A-Za-z0-9_. ()-]+$")
```

A little easier to read without all the escapes. (Note the `-` is now at the end so it will be interpreted literally.)

austinbyers (commented):
Deferring final approval to @nhakmiller , but looks good to me!

For context, we ban those characters to prevent HTML injection attacks

nhakmiller (commented):
Looks good! Love this clean implementation. Could you add a unit test (in `tests/unit/panther_analysis_tool`) to check for this case? You can add new test cases in `test/fixtures` as needed.

wey-chiang:
> Could you add a unit test (in `tests/unit/panther_analysis_tool`) to check for this case? 

Sure! I've added `example_policy_invalid_characters.py` and `example_policy_invalid_characters.yml` to the fixtures dir

wey-chiang:
> Looks good! Love this clean implementation. Could you add a unit test (in `tests/unit/panther_analysis_tool`) to check for this case? You can add new test cases in `test/fixtures` as needed.

Sure! I've added `example_policy_invalid_characters.py` and `example_policy_invalid_characters.yml` to the fixtures dir and updated `unit/panther_analysis_tool/test_main.py` accordingly.

Note that for `test_invalid_rule_definition` the `assert_equal(len(invalid_specs), 1)` statement had to be incremented e.g. `assert_equal(len(invalid_specs), 2)` with the new addition of the `example_policy_invalid_characters`

nhakmiller (approved):
LGTM

PR #48, opened 2020-09-24
Update to yaml verification

### LINKED ISSUE #32: Inadequate YAML Verification
The `panther_analysis_tool` is not validating the YAML files as rigorously as the Panther backend is. This means that the `panther_analysis_tool` might parse a file as valid YAML and have it pass all tests, whereas upon uploading the rule/policy to Panther an error will be returned indicating that the YAML is invalid.

When uploading, the error takes the following form:

```
 Panther CLI Test Summary
 	Path: rules/
 	Passed: 1
 	Failed: 0
 	Invalid: 0
[WARNING]: Failed to upload to Panther
 	status code: 400
 	error message: {"message":"yaml: line 18: did not find expected ',' or '}'"}
```

One user reported this error was observed with the following YAML being the cause:

```
    Log: 
      {
-        actor:{
+        actor: {
```

In this case, I think the Panther backend may be acting overly sensitively. However, the `panther_analysis_tool` should not allow YAML/JSON to pass verification that would not pass on the backend. We should more strictly check the YAML/JSON specification files to ensure they match exactly the format the backend is expecting.

lindsey-w:
### Background

The `panther_analsys_tool` isn't validating yaml as vigorously as the panther backed. We should add invalid yaml files to the invalid_specs when running `panther_analysis_tool test` Closes #32 

### Changes

* Added exception handling to the json.loads and yaml.safe_load calls so that any yaml or json file that fails to load will be added to the `invalid_specs` file

### Testing

* Added new invalidly formatted yaml file `example_malformed_yaml.yml` and updated tests to include additional invalid spec  
* `make test`

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -28,10 +28,10 @@ import os
 import sys
 import zipfile
 
+from ruamel.yaml import YAML, parser as YAMLParser, scanner as YAMLScanner
 from schema import (Optional, SchemaError, SchemaMissingKeyError,
                     SchemaForbiddenKeyError, SchemaUnexpectedTypeError)
 import boto3
-from ruamel.yaml import YAML, parser as YAMLParser, scanner as YAMLScanner
 
 from panther_analysis_tool.schemas import TYPE_SCHEMA, GLOBAL_SCHEMA, POLICY_SCHEMA, RULE_SCHEMA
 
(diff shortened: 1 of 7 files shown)


lindsey-w:
While I have updated this so that errors thrown by json.loads/yaml.safe_load will be handled more smoothly, I am still running into an issue with pyYAML.  Specifically, pyYAML will successfully read in a yaml doc that does not seem to conform to yaml, specifically when there isn't a space after a colon like: 

```
    Log:
      {
        UserName:{ 
          test: test
        }
      }
```

Still looking into this issue, as I haven't come up with a 'clean' way to handle it yet.

panther_analysis_tool/main.py
```diff
@@ -159,9 +171,10 @@ def zip_analysis(args: argparse.Namespace) -> Tuple[int, str]:
         # Always zip the helpers
         analysis = []
         files: Set[str] = set()
-        for (file_name, f_path, spec) in list(load_analysis_specs(
-                args.path)) + list(load_analysis_specs(HELPERS_LOCATION)):
-            if file_name not in files:
+        for (file_name, f_path, spec,
+             error) in list(load_analysis_specs(args.path)) + list(
+                 load_analysis_specs(HELPERS_LOCATION)):
+            if file_name not in files and not error:
```
nhakmiller, line 177: It should be extremely unlikely for this to ever return an error, because if the running `test_analysis` encounters any errors on lines 159-162 we exit before getting to this point correct? Are there specific cases you encountered or you're anticipating with this error check?

nhakmiller (approved):
One small question, non-blocking.

panther_analysis_tool/main.py
lindsey-w: That's a good point. I hadn't had a particular case in mind, I can remove it since its a duplicate error check

PR #50, opened 2020-09-29
test files current dir

### LINKED ISSUE #41: Cannot scan rules/policies in current directory
When using the `panther_analysis_tool`, passing a `--path` argument of `.` (or not specifying `--path`, which defaults to `.`) will skip in rules or policies in the current directory, only recursing into any sub-directories instead. This can be inconvenient when trying to test analysis because often you'll already be in the subdirectory for convenience of opening files etc.

lindsey-w:
### Background

This updates the directory walking logic so that running the `panther_analysis_tool` will test any policies/rules in the current dir as well as those within sub directories.  Closes #41 

### Changes

* update to directory walking structure
* new test case for `--path ./` and default (no `--path` option specified)

### Testing

* `make ci`

### DIFF
--- a/tests/unit/panther_analysis_tool/test_main.py
+++ b/tests/unit/panther_analysis_tool/test_main.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 import os
 
-from pyfakefs.fake_filesystem_unittest import TestCase
+from pyfakefs.fake_filesystem_unittest import TestCase, Pause
 from nose.tools import (assert_equal, assert_false, assert_is_instance,
                         assert_is_none, assert_true, raises, nottest,
                         with_setup)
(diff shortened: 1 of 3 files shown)


tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -46,6 +46,34 @@ def test_rules_from_folder(self):
         assert_equal(return_code, 0)
         assert_equal(len(invalid_specs), 0)
 
+    def test_rules_from_current_dir(self):
+        # This is a work around to test running tool against current directory
```
lindsey-w, line 50: A note about the testing method:
Using `pyfakefs` in conjunction with `importlib` causes some odd behavior that may be related to how things are patched.  I tried to add a test using the pyfakefs in place, but ran into errors.  Tests attempted:

1. Changing current working directing into sub dir within the fakefs and running the test from that context:
`os.chdir(/test/fixtures) -> run test -> os.chdir(original_working_dir)` This results in: “file or directory does not exist” for all the *.py files (seems to read the *.yml files)
2. Adding the files to the fakefs by using add_real_file  into the current working directory. This also results in: “file or directory does not exist” for the *.py files copied into the directory even though testing via `assert_true` (shown below) do not through an exception.  (again, seems to read the *.yml file)
```
self.fs.add_real_file(self.fixture_path+'example_policy.py',target_path="/example_policy.py")
self.fs.add_real_file(self.fixture_path+'example_policy.yml',target_path="/example_policy.yml")
assert_true(os.path.exists('./example_policy.yml'))
assert_true(os.path.exists('./example_policy.py'))
```
In both of these cases:
- os.listdir shows that the *.py files do “exist”
- The difference between reading the yml and the py file is that .yml is read in using open and the .py file is read in using importlib methods; which is why I thought perhaps an issue with pyfakefs patching methods. 

Worked around this by pausing the fakefs context to access the host filesystem, and resuming the fakefs for the remaining tests.

nhakmiller (approved):
Love the unit tests, thanks for adding this. I know it will help me a lot in my detection writing workflow.

PR #58, opened 2020-12-01
spec loading multiple times

lindsey-w:
### Background

The tool tries to be helpful by always including globals/data models, but this can cause them to be imported multiple times when running in current directory(imports both ./global_helpers and *global_helpers*) Because it imports multiple times and the latest release does more strict checking of analysis IDs - this will cause running `test` against the panther-analysis repo to fail.  

Update to upload payload format to take into account the new analysis-api

### Changes

* Check that each file imported is unique
* upload payload update, try old format if new format fails

### Testing

* `make ci` 
* run test in the panther-analysis repo: `python ../panther_analysis_tool/panther_analysis_tool/main.py test`
* confirmed upload to both panther v1.13.0 and v1.14.0 analysis-api works

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -272,7 +272,7 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
                     'statusCode') != 200:
                 logging.warning(
                     'Failed to upload to Panther\n\tstatus code: %s\n\terror message: %s',
-                    response_payload['statusCode'], response_payload['body'])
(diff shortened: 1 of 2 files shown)


nhakmiller (approved):

nhakmiller:
Should we do another release for this I suppose?

lindsey-w:
> Should we do another release for this I suppose?

Roger roger - updated versions, and I included the change to the upload payload. Let me know if you see something amiss or if it is still good to merge?

nhakmiller (approved):
LGTM

PR #62, opened 2020-12-22
Use Mapping for TestCase

jacknagz:
### Background

tl;dr the `event` passed around in local testing was not of type `dict` or `Mapping`, resulting in some problems. This fixes it.

When checking for types in helpers, I ran into an issue where the `event` was not of type `dict`. So after digging, I saw that we wrapped the events in a `TestCase` class that was not a subtype of `dict`, which broke the helper I wrote. When also digging through the rules engine, I saw that we subclass our [EnrichedEvent](https://github.com/panther-labs/panther/blob/master/internal/log_analysis/rules_engine/src/enriched_event.py#L24) from `Mapping`, which allows the subtypes to easily be inferred. 

### Changes

* Subclass `TestCase` from `Mapping`
* Bump version
* Update our README a little to remove unnecessary info

### Testing

I installed this version locally and validated my hypothesis by using a helper such as:

```python
def deep_get(dictionary, *keys, default=None):
    '''Safely return the value from a nested map'''
    return reduce(lambda d, key: d.get(key, default) if isinstance(d, Mapping) else default, keys, dictionary)
```

This now works as expected

Also ran the local unit tests to ensure I didn't break anything

### DIFF
--- a/panther_analysis_tool/test_case.py
+++ b/panther_analysis_tool/test_case.py
@@ -47,6 +47,9 @@ class TestCase(Mapping):
     def __len__(self) -> int:
         return len(self._data)
 
+    def get(self, arg: str, default: Any = None) -> Any:
+        return self._data.get(arg, default)
+
     def udm(self, key: str) -> Any:
         """Converts standard data model field to logtype field"""
         # ensure that rules using `udm` have included p_log_type in their test
--- a/panther_analysis_tool/test_case.py
+++ b/panther_analysis_tool/test_case.py
@@ -1,3 +1,4 @@
+from collections.abc import Mapping
 import logging
 from typing import Any, Callable, Dict, Iterator, List, Optional
 from jsonpath_ng import Fields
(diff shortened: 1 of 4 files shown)


panther_analysis_tool/test_case.py
```diff
@@ -43,8 +44,8 @@ def __getitem__(self, arg: str) -> Any:
     def __iter__(self) -> Iterator:
         return iter(self._data)
 
-    def get(self, arg: str, default: Any = None) -> Any:
-        return self._data.get(arg, default)
```
nhakmiller, line 47: I think we still need this function. I checked out this code and ran it against `panther-analysis` and it threw a bunch of errors without this.

jacknagz: Updated, thanks for checking that!

nhakmiller (approved):
Ok I tested the latest version and everything seems to be working fine. I'll merge & create the 0.4.3 release.

PR #63, opened 2021-01-08
hidden error message

lindsey-w:
### Background

In adding support for both the older and newer version of the `analysis-api` - if an error was returned from the new version it was hidden because it automatically re-tried the upload using the old format. This removes support for the old format, as users can choose which version of the `pat` to use based on what version of `panther` they have deployed. 

### Changes

* removed support for the old format

### Testing

* `make ci`
* manual testing


nhakmiller (approved):
We should remember to put a comment in the release notes about this.

wey-chiang (approved):
LGTM

PR #64, opened 2021-01-13
Continue if one global fails

nhakmiller:
### Background

Currently, if one global helper fails to load we stop loading any further global helpers. Instead, we should keep loading each global helper that we can load.

### Changes

* Keep loading global helpers even if one fails to load 

### Testing

* Tested in `panther-analysis` with one global that failed to load. Previous to this change multiple other globals were skipped, with this change only the failing global was skipped

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -610,7 +610,7 @@ def setup_parser() -> argparse.ArgumentParser:
         prog='panther_analysis_tool')
     parser.add_argument('--version',
                         action='version',
-                        version='panther_analysis_tool 0.4.3')
+                        version='panther_analysis_tool 0.4.4')
     subparsers = parser.add_subparsers()
 
     test_parser = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


lindsey-w (approved):
LGTM - thanks for finding/fixing 🙏🏻

panther_analysis_tool/main.py
```diff
@@ -322,7 +322,7 @@ def setup_global_helpers(global_analysis: List[Any]) -> List[Any]:
         # If the module could not be loaded, continue to the next
         if load_err:
             invalid_specs.append((analysis_spec_filename, load_err))
-            break
+            continue
```
wey-chiang, line 325: Should we be logging a message here?

wey-chiang (approved):
LGTM, left a question.

panther_analysis_tool/main.py
nhakmiller: We add the filename and err to the `invalid_specs` dict that gets passed around and eventually printed at the end so it should be logged eventually already.

PR #66, opened 2021-01-21
Add contains function

nhakmiller:
### Background

Support use of the `in` keyword for unit tests.

### Changes

* Add python `__contains__` magic function

### Testing

* Wrote a unit tests that uses `in` in a rule


lindsey-w (approved):
LGTM

PR #71, opened 2021-02-10
Add subcommand to update managed schemas

alxarch:
### Background

Closes [this issue](https://github.com/panther-labs/panther/issues/2639) 
We want to implement a CLI way to update managed schemas until FE team picks up the task to expose the functionality in the UI.

### Changes

* Adds `update-schemas` sub-command to fetch the available updates and choose the version to install

### Testing

* Ran the task against a deployed Panther version

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -285,37 +285,52 @@ def update_schemas(args: argparse.Namespace) -> Tuple[int, str]:
                                  {'ListManagedSchemaUpdates': {}}))
     response_str = response['Payload'].read().decode('utf-8')
     response_payload = json.loads(response_str)
-    api_err = response_payload['error']
+
+    api_err = response_payload.get('error')
     if api_err is not None:
         logging.error(
             'Failed to list managed schema updates\n\tcode: %s\n\terror message: %s',
             api_err['code'], api_err['message'])
         return 1, ''
-    releases = response_payload['releases']
-    if releases is None:
-        logging.Info('No updates available.')
+
+    releases = response_payload.get('releases')
+    if releases is None or len(releases) is 0:
+        logging.info('No updates available.')
         return 0, ''
-    print('Available versions:')
-    for (i, release) in enumerate(releases):
-        print('%d: %s' % (i, release.get('tag')))
-    choice = int(input('choose which version to install: '))
-    chosen = releases[choice]
+
+    tags = [r.get('tag') for r in releases]
+    tagLatest = tags[-1]
+    while True:
+        print('Available versions:')
+        for tag in tags:
+            print('\t%s' % tag)
+        print('Panther will update managed schemas to the latest version (%s)' %
+              tag)
+        prompt = 'Choose a different version (%s): '.format(tag)
+
+        choice = input(prompt).strip() or tagLatest  # nosec
+        if choice in tags:
+            break
+        else:
+            logging.error('Chosen tag %s is not valid', choice)
+
+    manifestURL = releases[tags.index(choice)].get('manifestURL')
 
     response = client.invoke(FunctionName='panther-logtypes-api',
                              InvocationType='RequestResponse',
                              Payload=json.dumps({
                                  'UpdateManagedSchemas': {
-                                     'release': chosen.get('tag'),
-                                     'manifestURL': chosen.get('manifestURL')
+                                     'release': choice,
+                                     'manifestURL': manifestURL
                                  }
                              }))
     response_str = response['Payload'].read().decode('utf-8')
     response_payload = json.loads(response_str)
-    api_err = response_payload['error']
+    api_err = response_payload.get('error')
     if api_err is not None:
         logging.error(
             'Failed to submit managed schema update to %s\n\tcode: %s\n\terror message: %s',
-            chosen.get('tag'), api_err['code'], api_err['message'])
+            choice, api_err['code'], api_err['message'])
         return 1, ''
     logging.info('Managed schemas updated successfully')
     return 0, ''
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -87,8 +87,8 @@ def load_module(filename: str) -> Tuple[Any, Any]:
     return module, None
 
 
-def load_analysis_specs(
-        directories: List[str]) -> Iterator[Tuple[str, str, Any, Any]]:
+def load_analysis_specs(directories: List[str]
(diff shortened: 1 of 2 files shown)


panther_analysis_tool/main.py
```diff
@@ -758,9 +841,10 @@ def run() -> None:
     logging.basicConfig(format='[%(levelname)s]: %(message)s',
                         level=logging.DEBUG if args.debug else logging.INFO)
 
+    if hasattr(args, 'filter') and args.filter is not None:
```
georgepsarakis, line 844: Minor, should be equivalent to `if getattr(args, 'filter', None) is not None`.

panther_analysis_tool/main.py
```diff
@@ -259,6 +259,83 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+    api_err = response_payload.get('error')
+    if api_err is not None:
+        logging.error(
+            'Failed to list managed schema updates\n\tcode: %s\n\terror message: %s',
+            api_err['code'], api_err['message'])
+        return 1, ''
+
+    releases = response_payload.get('releases')
+    if releases is None or len(releases) is 0:
```
georgepsarakis, line 297: Equivalent: `if not releases`

Note that `is` checks for object identity not equality. The check here happens to be correct, because up to a certain number integer instances are cached. For large numbers this isn't true:

```python
>>> id(0)
4454249536
>>> id(0)
4454249536
>>> id(0)
4454249536
>>> id(93939390303030303030303003030303030303003)
4456905776
>>> id(93939390303030303030303003030303030303003)
4456905728
>>> id(93939390303030303030303003030303030303003)
4456905776
```

panther_analysis_tool/main.py
```diff
@@ -259,6 +259,83 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+        logging.info('No updates available.')
+        return 0, ''
+
+    tags = [r.get('tag') for r in releases]
```
georgepsarakis, line 301: Is it possible that they key `tag` is missing? I would say better have an exception be raised, rather than end up with `None` values, which may also break line 317.

panther_analysis_tool/main.py
```diff
@@ -259,6 +259,83 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+    tagLatest = tags[-1]
```
georgepsarakis, line 302: `tagLatest -> latest_tag` in Python 😄

panther_analysis_tool/main.py
```diff
@@ -259,6 +259,83 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+              tag)
+        prompt = 'Choose a different version (%s): '.format(tag)
+
+        choice = input(prompt).strip() or tagLatest  # nosec
+        if choice in tags:
+            break
+        else:
+            logging.error('Chosen tag %s is not valid', choice)
+
+    manifestURL = releases[tags.index(choice)].get('manifestURL')
```
georgepsarakis, line 317: `manifestURL -> manifest_url`

georgepsarakis: Can the Lambda invocation handle `None` values? Maybe we could log an error here and fail.

panther_analysis_tool/main.py
alxarch: ah yeah

panther_analysis_tool/main.py
alxarch: These are always there in the response if the element exists.

alxarch: The worse case scenario would be that an empty manifest URL results in a Lambda error, but there's code to explicitly only return the entries with a valid `manifestURL` in the `ListManagedSchemaUpdates` endpoint.

panther_analysis_tool/main.py
alxarch: There's no such possibility. Release feed is filtering out any such problematic cases. Both fields are `required`

panther_analysis_tool/main.py
alxarch: The way I understand this would be that I check that `relates` is not `null` in the JSON and that it is not `[]` either.
Is this the same as `if not releases:`? How weird that there's a boolean value for lists. Forgot _that_ bug generator existed in Python.

panther_analysis_tool/main.py
alxarch: fixed

panther_analysis_tool/main.py
alxarch: fixed

panther_analysis_tool/main.py
alxarch: fixed

panther_analysis_tool/main.py
alxarch: fixed

panther_analysis_tool/main.py
georgepsarakis: As far as I remember all empty objects (sets, tuples, dicts, lists, strings) have a truth value of `False`.

panther_analysis_tool/main.py
```diff
@@ -259,6 +259,84 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+    tags = [r.get('tag') for r in releases]
+    latest_tag = tags[-1]
+    while True:
+        print('Available versions:')
+        for tag in tags:
+            print('\t%s' % tag)
+            print(
+                'Panther will update managed schemas to the latest version (%s)'
+                % tag)
```
nhakmiller, line 309: Should this be unindented one tab? It seems this will print this line after every single available version.

nhakmiller (approved):

panther_analysis_tool/main.py
wey-chiang: @georgepsarakis hit the nail on the head

Python will evaluate `if __OBJECT__` as bool `False` if `__OBJECT__` is the default value or None

panther_analysis_tool/main.py
georgepsarakis: True. I think it was like this initially, probably a code reformat changed it ...

georgepsarakis (approved):
👍

panther_analysis_tool/main.py
alxarch: Fixed, this originally correct but I mistakenly nudged it in during `make fmt` issues

PR #72, opened 2021-02-10
Update license to AGPLv3

s0l0ist:
### Background

Change the license to AGPLv3

### Changes

* <Describe changes here>

### Testing

* `pip-licenses` reflects the change


nhakmiller (approved):
LGTM, I suspect we should announce this change somewhere though.

wey-chiang (approved):
I agree w/ Nick H. as we should announce this change somewhere.

LGTM

PR #75, opened 2021-03-01
Add zip-schemas command to Panther analysis tool to streamline the release process

alxarch:
### Background

The release process should be as simple as possible.
Adding the code to package the managed schemas in PAT will help with that.
The `zip-schemas` sub-command does a shallow clone of a specific tag from the `panther-analysis` repository and packages a `manifest.yml` into a `.zip` archive to be attached to releases.  

#### Usage
```sh
$ panther_analysis_tool zip-schemas --release=v1.16.0
$ panther_analysis_tool zip-schemas --release=v1.16.0 --out=/tmp/
```
`--release` is a required argument for the release tag to package.
`--out` sets the output _directory_ (defaults to `.`).

### Changes

* Adds `zip-schemas`  command.

### Testing

* ran `zip-schemas`

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -843,7 +843,8 @@ def zip_managed_schemas(args: argparse.Namespace) -> Tuple[int, str]:
         repo_url = "https://github.com/panther-labs/panther-analysis"
         repo_dir = os.path.join(tmp_dir, "panther-analysis")
 
-        logging.info("Cloning %s tag of %s %s", args.release, repo_url, repo_dir)
+        logging.info("Cloning %s tag of %s %s", args.release, repo_url,
+                     repo_dir)
         result = subprocess.run([
             "git", "clone", "--branch", args.release, "--depth", "1", "-c",
             "advice.detachedHead=false", repo_url, repo_dir
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -862,14 +862,14 @@ def zip_managed_schemas(args: argparse.Namespace) -> Tuple[int, str]:
                           args.release)
             return 1, ""
 
-        logging.info('Building manifest.yml for %d managed schemas found in release %s',
-                     len(filenames), args.release)
+        logging.info(
+            'Building manifest.yml for %d managed schemas found in release %s',
+            len(filenames), args.release)
         for filename in filenames:
             with open(filename) as f:
                 lines = f.readlines()
                 manifest.append("---\n")
                 manifest.extend(lines)
-
 
     archive = os.path.join(args.out,
                            'managed-schemas-{}.zip'.format(args.release))
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -17,23 +17,24 @@ You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 '''
 
-from collections import defaultdict
-import subprocess
-import tempfile
-from datetime import datetime
-from fnmatch import fnmatch
-from importlib.abc import Loader
-from typing import Any, DefaultDict, Dict, Iterator, List, Set, Tuple
-import argparse
-import base64
-import importlib.util
 import json
 import logging
 import os
 import re
 import sys
+import subprocess  # nosec
 import zipfile
+import argparse
+import base64
+import importlib.util
+import tempfile
+from collections import defaultdict
+from datetime import datetime
+from fnmatch import fnmatch
+from importlib.abc import Loader
+from typing import Any, DefaultDict, Dict, Iterator, List, Set, Tuple
 
+import semver
 from ruamel.yaml import YAML, parser as YAMLParser, scanner as YAMLScanner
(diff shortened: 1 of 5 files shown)


nhakmiller (commented):
Seems mostly fine, but can you add a unit test to `tests/unit/panther_analysis_tool/test_main.py` for this function? We have a basic one for the existing `zip` command you can copy & tweak.

lindsey-w:
Wondering if there is a reason to pull from github for the schemas vs using what is in the local dir (which is what we do for detections)? I'm guessing it has to do with the versioning of the schemas vs the rules/policies aren't tied as much to versions?

panther_analysis_tool/main.py
```diff
@@ -810,6 +828,61 @@ def setup_parser() -> argparse.ArgumentParser:
     return parser
 
 
+def zip_managed_schemas(args: argparse.Namespace) -> Tuple[int, str]:
```
lindsey-w, line 831: Is the proposed process to: 1. create a release in github 2. after the release is created, run `pat` to generate the assets? 3. update the release to include the assets?

alxarch: Yes, the `vX.Y.Z` tag must exist to produce the asset.

alxarch: If we ever automate this we just need to do the `git tag` before we issue a 'create release' git hub api call.

alxarch:
> Seems mostly fine, but can you add a unit test to `tests/unit/panther_analysis_tool/test_main.py` for this function? We have a basic one for the existing `zip` command you can copy & tweak.

I tried but failed because the `self.fs` propery seems to be some weird mock thing that does not read from an actual filesystem. So the test fails while the actual execution succeeds.

panther_analysis_tool/main.py
```diff
@@ -810,6 +828,61 @@ def setup_parser() -> argparse.ArgumentParser:
...
+    Returns:
+        A tuple of return code and the archive filename.
+    """
+
+    manifest = []
+    with tempfile.TemporaryDirectory(prefix="zip-managed-schemas-") as tmp_dir:
+        repo_url = "https://github.com/panther-labs/panther-analysis"
+        repo_dir = os.path.join(tmp_dir, "panther-analysis")
+
+        logging.info("Cloning %s tag of %s", args.release, repo_url, repo_dir)
```
lindsey-w, line 846: is `repo_dir` used here?

panther_analysis_tool/main.py
```diff
@@ -810,6 +828,61 @@ def setup_parser() -> argparse.ArgumentParser:
...
+        result = subprocess.run([
+            "git", "clone", "--branch", args.release, "--depth", "1", "-c",
+            "advice.detachedHead=false", repo_url, repo_dir
+        ])
```
lindsey-w, line 850: checking that args.release is a sem ver might fix the ci error

lindsey-w (commented):
Looks good to me - just failing a security check in ci.

nhakmiller (approved):
Looks good to me, I pushed up two small changes so the linter wouldn't complain so much. It seems the code checker still doesn't like the use of  subprocess though, but I'll leave it up to you to resolve.

panther_analysis_tool/main.py
alxarch: The error is confusing and of any value IMO.
I added a `check=True` to raise an exception on non-zero return code and a `timeout=120` to raise an exception if a process gets struck.
I resorted to `#nosec` since the only way to do this is by spawning `git`.

panther_analysis_tool/main.py
alxarch: fixed

alxarch:
> It seems the code checker still doesn't like the use of subprocess though, but I'll leave it up to you to resolve.

I added a `check=True` to raise an exception on non-zero return code and a `timeout=120` to raise an exception if a process gets struck.
I resorted to `#nosec` since the only way to do this is by spawning `git`.
I will let the merging to @nhakmiller so that he gets a chance to review the `# nosec` added.
At this point the security concern AFAICT is that `git` might resolve to a malicious tool. This to me sounds like paranoia.

panther_analysis_tool/main.py
```diff
@@ -810,6 +829,67 @@ def setup_parser() -> argparse.ArgumentParser:
...
+            logging.error("Invalid release tag %s", rel)
+            return 1, ""
+
+        logging.info("Cloning %s tag of %s", rel, repo_url)
+        # nosec
+        cmd = [
+            "git", "clone", "--branch", rel, "--depth", "1", "-c",
+            "advice.detachedHead=false", repo_url, repo_dir
+        ]
+        result = subprocess.run(cmd, check=True, timeout=120)  # nosec
```
georgepsarakis, line 858: `check=True` will raise an exception on non-zero exit codes, so the check below is not necessary, or better phrased this code branch will never run. Was [this](https://bandit.readthedocs.io/en/latest/_modules/bandit/plugins/injection_shell.html) the check you want to bypass?

alxarch:
There were 3 different errors all relating to subprocess.
One was complaining on the import itself.
One about passing `git` without a path to the executable.
One about using using subprocess.run

PR #79, opened 2021-03-09
Allow Advanced PAT Users to Ignore Extra Fields

wey-chiang:
### References

Asana Task URL: https://app.asana.com/0/1199975676341842/1199956661047416/f

### Background

Right now if you add an extraneous field to a detection specification file, PAT will fail to load that file considering it invalid.  This is to prevent users from accidentally mistyping an optional field and having it be dropped silently. 

It would be nice if this functionality could be overwritten so that users can add custom meta data fields which don't get translated over to Panther but which can contain useful information about the detection for a given organization.

### Changes

* Hacky solution to allow for configurable `ignore_extra_fields` flag

### Testing

* Tested `zip` `test` and `upload` with a rule specification that was not expected

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -33,10 +33,9 @@ from datetime import datetime
 from fnmatch import fnmatch
 from importlib.abc import Loader
 from typing import Any, DefaultDict, Dict, Iterator, List, Set, Tuple
-
+from distutils.util import strtobool
 import boto3
 import semver
-from distutils.util import strtobool
 from ruamel.yaml import YAML
 from ruamel.yaml import parser as YAMLParser
 from ruamel.yaml import scanner as YAMLScanner
@@ -965,7 +964,7 @@ def run() -> None:
     if getattr(args, "filter", None) is not None:
         args.filter = parse_filter(args.filter)
 
-    global RULE_SCHEMA, POLICY_SCHEMA
+    global RULE_SCHEMA, POLICY_SCHEMA  # pylint: disable=global-variable-undefined
     RULE_SCHEMA, POLICY_SCHEMA = get_rule_policy_schema(bool(args.ignore_extra_keys))
 
     try:
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -982,6 +982,7 @@ def run() -> None:
         level=logging.DEBUG if args.debug else logging.INFO,
     )
 
+    args.filter = None
     if getattr(args, "filter", None) is not None:
         args.filter = parse_filter(args.filter_list)
 
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -972,6 +972,11 @@ def set_env(key: str, value: str) -> None:
     os.environ[key] = value
 
 
+def set_rule_policy_schema(args: argparse.Namespace) -> None:
+    global RULE_SCHEMA, POLICY_SCHEMA  # pylint: disable=global-statement
+    RULE_SCHEMA, POLICY_SCHEMA = get_rule_policy_schema(bool(args.ignore_extra_keys))
+
+
 def run() -> None:
     parser = setup_parser()
     # if no args are passed, print the help output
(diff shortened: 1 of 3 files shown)


wey-chiang:
... idk what happened but it seems like the pyfakefs issues in the nosetests fixed itself.

panther_analysis_tool/schemas.py
```diff
@@ -57,68 +58,73 @@
...
-                    "LogType"
-                ): str,  # Not needed anymore, optional for backwards compatibility
-                "ExpectedResult": bool,
-                "Log": object,
-            }
-        ],
-    },
-    ignore_extra_keys=False,
-)  # Prevent user typos on optional fields
+def get_rule_policy_schema(ignore_extra_keys: bool = False) -> Tuple[Schema, Schema]:
```
nhakmiller, line 62: I see, this is the workaround you were talking about? It is a bit heavy.

I know it's not python best practice, but it seems we can just refer to the internal `_ignore_extra_keys` field to set this value. Little example I whipped up in an interpreter:

```python
>>> from schema import Schema
>>> thing = Schema({"Type": str,}, ignore_extra_keys=True)
>>> thing.validate({'Type': 'hello', 'Other': 'Meme'})
{'Type': 'hello'}
>>> thing._ignore_extra_keys  = False
>>> thing.validate({'Type': 'hello', 'Other': 'Meme'})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/schema.py", line 415, in validate
    raise SchemaWrongKeyError(message, e)
schema.SchemaWrongKeyError: Wrong key 'Other' in {'Type': 'hello', 'Other': 'Meme'}
```

So it seems we could leave this mostly the way it was before, and when the `--ignore-extra-fields` flag is set we can update the `POLICY_SCHEMA._ignore_extra_keys` variable. The `_` prefix means by convention we shouldn't interact with this directly. But for such a simple flag it seems almost like an oversight from the library to not make it modifiable, not that modifying it will have some strange behavior. Interested to hear what you think.

panther_analysis_tool/main.py
```diff
@@ -16,6 +16,7 @@
 You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+#  pylint: disable=too-many-lines
```
lindsey-w, line 19: I'm realizing that the PR I merged yesterday might cause some "fun" merge conflicts here. 😕

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -38,17 +39,19 @@ def test_handle_wrong_key_error(self):
...
         # test failing regex match
         test_str = "Will not match"
-        exc = Exception(test_str)
+        exc = SchemaWrongKeyError(test_str)
         err = pat.handle_wrong_key_error(exc, sample_keys)
         assert_equal(str(err),  expected_output.format("UNKNOWN_KEY", sample_keys))
 
     def test_load_policy_specs_from_folder(self):
         args = pat.setup_parser().parse_args('test --path tests/fixtures'.split())
+        args.filter = None
```
lindsey-w, line 53: Curious why this is required?

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -114,55 +125,65 @@ def test_aws_profiles(self):
...
         assert_equal(len(invalid_specs), 4)
 
     def test_invalid_characters(self):
         args = pat.setup_parser().parse_args('test --path tests/fixtures --filter Severity=High PolicyID=AWS.IAM.MFAEnabled'.split())
-        args.filter = pat.parse_filter(args.filter)
+        pat.set_rule_policy_schema(args)
+        args.filter = pat.parse_filter(args.filter_list)
         return_code, invalid_specs = pat.test_analysis(args)
-        assert_equal(return_code, 1)
+        self.equal = assert_equal(return_code, 1)
```
lindsey-w, line 147: Should this be?

```suggestion
        assert_equal(return_code, 1)
```

tests/unit/panther_analysis_tool/test_main.py
wey-chiang: I added the `dest` parameter to the `subparser.add_argument` function that maps the user input to `filter_list`
This was done for type resolving since we were remapping `filter` from a list to a dict which complicated the typing

wey-chiang: The reason it is required is because `args.filter` is not instantiated by `parse_args` now, it is now instantiated by the line `args.filter = parse_filter(args.filter_list)`

panther_analysis_tool/schemas.py
wey-chiang: That's doable, it'll throw a linting issue but we could ignore it.

wey-chiang:
Will rework this as there is a much cleaner way to implement this feature

wey-chiang:
Reworked PR here: https://github.com/panther-labs/panther_analysis_tool/pull/81

PR #81, opened 2021-03-10
Allow Advanced PAT Users to Ignore Extra Fields

wey-chiang:
### References

Asana Task URL: https://app.asana.com/0/1199975676341842/1199956661047416/f

### Background

Right now if you add an extraneous field to a detection specification file, PAT will fail to load that file considering it invalid.  This is to prevent users from accidentally mistyping an optional field and having it be dropped silently. 

It would be nice if this functionality could be overwritten so that users can add custom meta data fields which don't get translated over to Panther but which can contain useful information about the detection for a given organization.

### Changes

* Hacky solution to allow for configurable `ignore_extra_fields` flag

### Testing

* Tested `zip` `test` and `upload` with a rule specification that was not expected

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -1018,6 +1018,7 @@ def run() -> None:
     if getattr(args, "filter", None) is not None:
         args.filter = parse_filter(args.filter)
 
+    # Although not best practice, the alternative is rather ugly and significantly harder to maintain.
     if bool(getattr(args, "ignore_extra_keys", None)):
         RULE_SCHEMA._ignore_extra_keys = True  # pylint: disable=protected-access
         POLICY_SCHEMA._ignore_extra_keys = True  # pylint: disable=protected-access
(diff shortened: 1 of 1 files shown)


panther_analysis_tool/main.py
```diff
@@ -1007,6 +1018,10 @@ def run() -> None:
     if getattr(args, "filter", None) is not None:
         args.filter = parse_filter(args.filter)
 
+    if bool(getattr(args, "ignore_extra_keys", None)):
+        RULE_SCHEMA._ignore_extra_keys = True  # pylint: disable=protected-access
+        POLICY_SCHEMA._ignore_extra_keys = True  # pylint: disable=protected-access
```
nhakmiller, line 1023: A brief comment explaining why we went this route (basically that the alternative is super ugly) will help future maintainers of this code I think.

nhakmiller (approved):
One minor comment, otherwise LGTM

panther_analysis_tool/main.py
```diff
@@ -1007,6 +1018,10 @@ def run() -> None:
     if getattr(args, "filter", None) is not None:
         args.filter = parse_filter(args.filter)
 
+    if bool(getattr(args, "ignore_extra_keys", None)):
```
wey-chiang, line 1021:
```suggestion
    # Although not best practice, the alternative is rather ugly and significantly harder to maintain.
    if bool(getattr(args, "ignore_extra_keys", None)):
```

PR #85, opened 2021-03-22
Support for Mocking in Unit Tests & Generated Alert Fields

wey-chiang:
### Background
Asana URL: Closes https://app.asana.com/0/1199940726973098/1199936555115825/f

Mocking Support within Unit Tests have been added in Panther's Rules Engine. 
Similarly, generated alert fields are also fully supported in Panther's Rules Engine.

### Changes

* Added support for mocking in unit tests
* Added support for generated alert fields (with type checking)
* General housekeeping/refactoring

### Testing

* `make test --path=./test/` (custom test files)
* `make unit`
* `make integration`

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -979,44 +979,54 @@ def _verify_test_run(
 
 
 def validate_outputs(function_name: str, function_output: Any) -> (bool, Any):
-    valid = True
-    out = function_output
+    # Defaults to valid and function output
+    # Invalidating criteria will overwrite these values
+    is_valid = True
+    output = function_output
     if function_name == "severity":
+        # Type checking for severity
         if not isinstance(function_output, str):
-            valid = False
+            is_valid = False
+        # Value Validation for severity
         if str(function_output).upper() not in VALID_SEVERITIES:
-            valid = False
-            out = AssertionError(
+            is_valid = False
+            output = AssertionError(
                 f'Expected [{function_name}] to be any of the following: '
                 f'{str(VALID_SEVERITIES)}, got [{str(function_output)}] instead.'
             )
     elif function_name == "destinations":
+        # Type checking for destinations
         if isinstance(function_output, list):
+            # Type checking for elements in list
+            # Note: we cannot perform value validation unless we were to query the outputs-api
             if not all(isinstance(x, str) for x in function_output):
+                # List contains non-string element, so enumerate the types present for error msg
                 list_types = set()
                 for each_item in function_output:
                     list_types.add(type(each_item))
-                valid = False
-                out = AttributeError(
+                is_valid = False
+                output = AttributeError(
                     f'Expected [{function_name}] to return a list of strings, '
                     f'got types [{list_types}] instead'
                 )
         else:
-            valid = False
+            is_valid = False
+    # For reserved functions besides severity and destinations, we expect a string output
     else:
-        valid = isinstance(function_output, str)
+        is_valid = isinstance(function_output, str)
 
-    if not valid and out == function_output:
+    # Invalid Types, used to generate a corresponding error message
+    if not is_valid and output == function_output:
         if function_name == "destinations":
             expected_type = "list"
         else:
             expected_type = "string"
 
-        out = AttributeError(
+        output = AttributeError(
             f'Expected [{function_name}] to return a [{expected_type}], '
             f'got [{type(function_output)}] instead'
         )
-    return valid, out
+    return is_valid, output
 
 
 def setup_parser() -> argparse.ArgumentParser:
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -36,7 +36,7 @@ from distutils.util import strtobool
 from fnmatch import fnmatch
 from importlib.abc import Loader
 from typing import Any, DefaultDict, Dict, Iterator, List, Set, Tuple
-from unittest.mock import patch, MagicMock
+from unittest.mock import MagicMock, patch
 
 import boto3
 import botocore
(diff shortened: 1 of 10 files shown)


lindsey-w (commented):
Wondering if we should also ad valid and invalid mocking examples the test cases?

panther_analysis_tool/main.py
```diff
@@ -864,13 +868,29 @@ def run_tests(
...
+                    result = analysis_funcs["run"](test_case)
+            else:
+                result = analysis_funcs["run"](test_case)
+        except AttributeError as err:
+            logging.warning("AttributeError: {%s}", err)
+            failed_tests[analysis.get("PolicyID") or analysis["RuleID"]].append(unit_test["Name"])
+            continue
         except KeyError as err:
             logging.warning("KeyError: {%s}", err)
             failed_tests[analysis.get("PolicyID") or analysis["RuleID"]].append(unit_test["Name"])
```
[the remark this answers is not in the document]
wey-chiang: Combined the two since they're functionally equivalent

panther_analysis_tool/main.py
```diff
@@ -84,6 +74,14 @@
 SCHEDULED_RULE = "scheduled_rule"
 RULE = "rule"
 
+CUSTOM_FUNCTIONS = [
```
nhakmiller, line 77: Maybe we should call these `RESERVED_FUNCTIONS` or something similar, since they have particular meanings and must return certain types of values if defined.

panther_analysis_tool/main.py
```diff
@@ -888,28 +908,54 @@ def run_tests(
...
+                    func_out_valid_type = False
+                    # severity value validation
+                    if func == "severity":
+                        if str(func_out).upper() not in VALID_SEVERITIES:
+                            func_out = AssertionError(
+                                f'Expected severity to be any of the following: '
+                                f'{str(VALID_SEVERITIES)}, got [{str(func_out)}] instead.'
+                            )
+                            test_result[func] = "FAIL"
+                            test_result["outcome"] = "FAIL"
```
nhakmiller, line 946: Not sure if this is better or worse, but I think you can do this 
```suggestion
                            test_result[func] = test_result["outcome"] = "FAIL"
```
to save a few lines here and everywhere this is repeated.

nhakmiller (commented):
Looks mostly good, I second Lindsey's suggestion that we consider breaking out this function into some smaller pieces. For example, we could put the first try/catch block that only runs the rule into a standalone function, then the big if statement that only validates the output into a standalone function, and that could help separate out some responsibilities.

The one thing I'm holding back approval on is some unit tests. Seems like this shouldn't be too hard to test and the mocking especially seems a bit complex so we might want some tests around that and the output validation.

wey-chiang:
For reference: example output from testing

image

wey-chiang:
Note that for the generated alert functions, we do not include those for the total number of passing/failing

I'm wondering if this should be changed since we're now doing value validation for severity and type checking for the rest of them

panther_analysis_tool/main.py
```diff
@@ -888,38 +926,97 @@ def run_tests(
...
-        and [x for x in analysis["Tests"] if not x["ExpectedResult"]]
-    ):
-        failed_tests[analysis.get("PolicyID") or analysis["RuleID"]].append(
-            "Insufficient test coverage: expected at least one positive and one negative test"
+        else:
+            valid = False
+    else:
+        valid = isinstance(function_output, str)
+
+    if not valid and out == function_output:
```
lindsey-w, line 1009: nit: could you rename or add a comment here (or at declaration) on what `out` represents / what this comparison is doing? It was a little hard to follow what is actually being compared here.

panther_analysis_tool/main.py
```diff
@@ -888,38 +926,97 @@ def run_tests(
...
+                else:
+                    func_out = analysis_funcs[func](test_case)
+                    if not func_out:
+                        test_run_result[func] = "FAIL"
+            except Exception as err:  # pylint: disable=broad-except
+                func_out = err
+                test_run_result[func] = "FAIL"
+
+            func_out_valid_type = True
+            # severity value validation
```
lindsey-w, line 967: is this comment up to date?

tests/fixtures/example_rule_invalid_mocks.yml
```diff
@@ -0,0 +1,32 @@
+AnalysisType: rule
+Filename: example_rule_invalid_mocks.py
+DisplayName: Example Rule Invalid Mock
```
lindsey-w, line 3: could you add description on what makes this invalid?

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -167,6 +169,47 @@ def test_unknown_exception(self):
...
+        for valid_invalid_outputs in [example_valid_outputs, example_invalid_outputs]:
+            for each_example in valid_invalid_outputs:
+                result = validate_outputs(each_example[0], each_example[1])
+                if invalid:
+                    assert_false(result[0])
+                    assert_false(result[1] == each_example[1])
+                else:
+                    assert_true(result[0])
+                    assert_equal(result[1], each_example[1])
+            invalid = True
```
lindsey-w, line 211: personal preference: I would take the outer `for` loop out and make two tests - just for clarity on what is actually being tested in the asserts.  Definitely not necessary, though.

lindsey-w (approved):
This is awesome 🤩   

Only comment is just adding comments for clarity in a few places.

nhakmiller:
> Note that for the generated alert functions, we do not include those for the total number of passing/failing
> 
> I'm wondering if this should be changed since we're now doing value validation for severity and type checking for the rest of them

I think we should count it as failing if it fails validation.

panther_analysis_tool/main.py
nhakmiller: Yeah this function seems to be doing the correct thing, I also had to re-read a few times to follow along though. I'd suggest either some more comments or maybe some renaming to make it a bit more clear what exactly is going on.

panther_analysis_tool/main.py
```diff
@@ -888,38 +926,97 @@ def run_tests(
...
         if unit_test["ExpectedResult"]:
-            for func in ["dedup", "title"]:
-                if analysis_funcs.get(func):
-                    if not analysis_funcs[func](test_case):
-                        test_result[func] = "FAIL"
-                        test_result["outcome"] = "FAIL"
+            verify_result = _verify_test_run(test_case, mock_methods, analysis_funcs)
+            for function_name in verify_result:
+                if verify_result[function_name] != "PASS":
+                    test_result[function_name] = test_result["outcome"] = "FAIL"
```
nhakmiller, line 940: Yeah right here I think this should fail the test
```suggestion
                    test_result[function_name] = test_result["outcome"] = "FAIL"
                    failed_tests[analysis.get("PolicyID") or analysis["RuleID"]].append(unit_test["Name"])
```

nhakmiller (approved):
Looks good overall, especially appreciate the unit tests. I would say that I think we need to fail the test iff the validation fails, even if strictly speaking the rule would still run in Panther.

panther_analysis_tool/main.py
wey-chiang: Doh, nice catch :)

PR #86, opened 2021-03-22
Update profile logic

nhakmiller:
### Background

Change how the AWS profile is set.

### Changes

* Change how the AWS profile is set

### Testing

* Local testing


lindsey-w (approved):
LGTM

PR #91, opened 2021-03-30
Support case-insensitive lookup for test cases

georgepsarakis:
### Background

Addresses case-sensitive lookup support: https://app.asana.com/0/1200124882238265/1200116099665098

### Changes

* `TestCase`: Add case-insensitive lookup partial support for top-level fields only.
* `TestCase`: Remove non-required overrides for the Mapping sub-class and use inherited behavior.

### Testing

* Unit tests
* Executed `panther_analysis_tool test` on the latest master branch of `panther-analysis`.

### DIFF
--- a/panther_analysis_tool/test_case.py
+++ b/panther_analysis_tool/test_case.py
@@ -35,12 +35,19 @@ class TestCase(Mapping):
         """
         self._data = data
         self.data_model = data_model
+        self._case_insensitive_key_map: Dict[str, str] = {}
+        # Support case-insensitive lookups for top-level fields only
+        for key in self._data:
+            self._case_insensitive_key_map.setdefault(key.lower(), key)
 
     def __getitem__(self, arg: str) -> Any:
-        return self._data.get(arg, None)
-
-    def __contains__(self, key: object) -> bool:
-        return key in self._data
+        try:
+            return self._data[arg]
+        except KeyError as key_error:
+            original_key = self._case_insensitive_key_map.get(arg.lower())
+            if original_key is None:
+                raise key_error
+            return self._data[original_key]
 
     def __iter__(self) -> Iterator:
         return iter(self._data)
@@ -48,8 +55,8 @@ class TestCase(Mapping):
     def __len__(self) -> int:
         return len(self._data)
 
-    def get(self, key: str, default: Optional[object] = None) -> Any:
-        return self._data.get(key, default)
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__}({self._data})'
 
     def udm(self, key: str) -> Any:
         """Converts standard data model field to logtype field"""
--- a/tests/unit/panther_analysis_tool/test_test_case.py
+++ b/tests/unit/panther_analysis_tool/test_test_case.py
@@ -1,0 +1,20 @@
+import unittest
+from panther_analysis_tool.test_case import TestCase as PantherTestCase
+
+
+class TestPantherTestCase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.data = {"username": "someone@company.com", "ip": "10.0.0.1"}
+        self.data_model = None
+        self.test_case = PantherTestCase(self.data, self.data_model)
+
+    def test_get(self) -> None:
+        self.assertEqual(self.test_case.get('username'), self.data['username'])
+
+    def test_missing_key_lookup(self) -> None:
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -21,3 +21,6 @@ jobs:
         - run:
             name: CLI Tests
             command: make ci && pipenv run make integration
+        - run:
+            name: Unit Tests
+            command: make unit
```
lindsey-w, line 26: Is this needed? I think `make ci` will run the unit tests. I think the it will also run the integration tests, so I'm not sure why the previous command is running those again either 🤔

georgepsarakis: Aha, you are probably right, I'll check it thanks. This is not ready for review yet, so don't waste your time, I am still trying out a few things, mostly to align pipenv - requirements.txt dependencies and workflow.

panther_analysis_tool/main.py
```diff
@@ -573,9 +573,18 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
     """
     logging.info("Testing analysis packs in %s\n", args.path)
 
+    search_directories = [args.path]
+
+    # Try the parent directory as well
```
georgepsarakis, line 578: I think this is necessary in order to allow #41.

dev-requirements.txt
```diff
@@ -0,0 +1,36 @@
+#
```
georgepsarakis, line 1: I can move the dependency versioning and related workflow modifications to a separate PR if you think that's more convenient.

panther_analysis_tool/main.py
```diff
@@ -573,9 +573,18 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
...
+    for directory in (HELPERS_LOCATION, '.' + HELPERS_LOCATION,
```
georgepsarakis, line 579: Are there cases that scanning the parent directory can cause issues, e.g. conflicts?

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -94,14 +95,16 @@ def test_rules_from_current_dir(self):
...
-        valid_rule_path = self.fixture_path + 'valid_analysis/policies'
+        valid_rule_path = os.path.join(FIXTURES_PATH, 'valid_analysis/policies')
         # test default path, '.'
         with Pause(self.fs):
             original_path = os.getcwd()
-            os.chdir(valid_rule_path)
-            args = pat.setup_parser().parse_args('test'.split())
-            return_code, invalid_specs = pat.test_analysis(args)
-            os.chdir(original_path)
+            try:
```
georgepsarakis, line 102: This is a safeguard, in case an exception occurs in the following lines, the directory would not be switched back, which is a potential state leak for other tests.

panther_analysis_tool/test_case.py
```diff
@@ -35,21 +35,28 @@ def __init__(self, data: Dict[str, Any], data_model: Optional[DataModel]) -> Non
         """
         self._data = data
         self.data_model = data_model
+        self._case_insensitive_key_map: Dict[str, str] = {}
+        # Support case-insensitive lookups for top-level fields only
+        for key in self._data:
+            self._case_insensitive_key_map.setdefault(key.lower(), key)
 
     def __getitem__(self, arg: str) -> Any:
-        return self._data.get(arg, None)
```
georgepsarakis, line 40: I took the liberty of modifying the behavior here as mentioned before, which also uncovered a small bug in a `panther-analysis` test.

georgepsarakis:
@lindsey-w this should be ready for review. As mentioned [here](https://github.com/panther-labs/panther_analysis_tool/pull/91#pullrequestreview-625965587), let me know if you want me to split this to two separate PRs. Thanks!

nhakmiller:
> As mentioned [here](https://github.com/panther-labs/panther_analysis_tool/pull/91#pullrequestreview-625965587), let me know if you want me to split this to two separate PRs. Thanks!

I do think it makes most sense to split these out into two PRs, as the aspects should really be considered separately and one can run without the other.

georgepsarakis:
Opened #92 with the unrelated changes and removed the commits from this branch.

lindsey-w (approved):
LGTM

nhakmiller (approved):

PR #92, opened 2021-04-02
Dependency Management Workflow Fixes

georgepsarakis:
### Background

`pipenv` and requirements-file based dependency management should use a consistent flow and direction.
Since `pipenv` can generate (and import) requirements files, we can perform all changes there and sync the requirements files during dependency version updates.

In order to have platform-agnostic environment setup, we also set the virtualenv construction and activation as the first step before installing dependencies (`pipenv` included).

The unit tests could also not be executed from within the unit tests directory, which facilitates debugging and we don't need to rely on specific current working directory for unit test execution.

Initially implemented as part of #91.

### Changes

* Align `setup.py` with `requirements.txt` dependency versions.
* Align `Pipfile` with `requirements.txt` dependency versions.
* Transition all dependency installation and update workflows to work with `pipenv`. Note that changes requirements files can still be imported. Requirements files (dev & default) are now auto-generated.
* Allow unit tests to be executed from any working directory (facilitates IDE debugging).
* Fall back to parent directory to check for global helpers and data models, if they exist. Enables running tests from the current working directory, which only contains rule / policy definitions.
* Use absolute paths when discovering and loading specification files.

### Testing

* CI
* Executed `panther_analysis_tool test` on the latest master branch of `panther-analysis`.


panther_analysis_tool/main.py
```diff
@@ -573,9 +573,18 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
     """
     logging.info("Testing analysis packs in %s\n", args.path)
 
+    search_directories = [args.path]
+
+    # Try the parent directory as well
+    for directory in (HELPERS_LOCATION, '.' + HELPERS_LOCATION,
+                DATA_MODEL_LOCATION, '.' + DATA_MODEL_LOCATION):
```
lindsey-w, line 580: This is nice. I don't think there would be any conflicts.

lindsey-w (commented):
Seems like a lot of nice improvements.

nhakmiller (approved):

PR #94, opened 2021-04-07
Use Python dict for policy test input

georgepsarakis:
### Background

https://github.com/panther-labs/panther_analysis_tool/pull/91 introduces incompatible changes between rule and policy processing. Using a plain `dict` as input for policies replicates what also happens in production.

### Changes

* Use `dict` instead of `TestCase` object when running policy tests.
* Print stack traces when using `--debug` in order to allow easier debugging.

### Testing

* Ran tests on current master branch of `panther-analysis`. When an error is raised and `--debug` flag is used, the stack trace is also printed.

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -912,8 +912,12 @@ def _run_tests(
                     for each_mock in mocks
                     if "objectName" in each_mock and "returnValue" in each_mock
                 }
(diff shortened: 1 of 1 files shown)


lindsey-w (approved):

georgepsarakis:
Thanks for reviewing @lindsey-w !

PR #96, opened 2021-04-22
Support for Log Types & Resource Types Validation

wey-chiang:
Closes https://app.asana.com/0/1200175800812961/1199956661047407/f

### Background

The Panther backend validates the `LogTypes` and `DetectionTypes` fields, however we currently do not support this validation in `panther_analysis_tool`. This PR adds support for validating these fields to provide users more consistency and advanced warning when using `panther_analysis_tool`.

Note that any additions in Panther necessitate changes here as well -- another case where we need some type of shared libraries between the two repositories

### Changes

* Added support via Regex for LogTypes and ResourceTypes
* Added flags that allow a user to ignore these validations for cases where there are custom LogTypes and ResourceTypes present
* Fixed typo `repsitory` -> `repository`

### Testing

* `make test`
* Tested in local environment, see attached SS
image

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -1213,13 +1213,13 @@ def setup_parser() -> argparse.ArgumentParser:
     )
     publish_parser.add_argument(
         "--github-owner",
-        help="The github owner of the repsitory",
+        help="The github owner of the repository",
         type=str,
         default="panther-labs",
     )
     publish_parser.add_argument(
         "--github-repository",
-        help="The github repsitory name",
+        help="The github repository name",
         type=str,
         default="panther-analysis",
     )
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -776,7 +776,7 @@ def classify_analysis(
     specs: List[Tuple[str, str, Any, Any]],
     ignore_logtypes_validation: bool = False,
     ignore_resourcetypes_validation: bool = False,
-) -> Tuple[Dict[str, List[Any]], List[Any]]:
+) -> Tuple[Dict[str, List[Any]], List[Any]]:  # pylint: disable=too-many-locals,too-many-statements
 
     # First setup return dict containing different
     # types of detections, meta types that can be zipped
@@ -790,6 +790,7 @@ def classify_analysis(
     # add any duplicates to the invalid_specs
     analysis_ids: List[Any] = []
 
+    # pylint: disable=too-many-nested-blocks
     for analysis_spec_filename, dir_name, analysis_spec, error in specs:
         keys: List[Any] = list()
         tmp_logtypes: Any = None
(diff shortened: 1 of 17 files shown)


panther_analysis_tool/schemas.py
```diff
@@ -20,6 +20,35 @@
...
+    r"|Fluentd\.Syslog5424|GCP\.AuditLog|GitLab\.API|GitLab\.Audit|GitLab\.Exceptions|GitLab\.Git"
+    r"|GitLab\.Integrations|GitLab\.Production|Gravitational\.TeleportAudit|GSuite\.Reports"
+    r"|Juniper\.Access|Juniper\.Audit|Juniper\.Firewall|Juniper\.MWS|Juniper\.Postgres|Juniper"
+    r"\.Security|Lacework\.Events|Microsoft365\.Audit\.AzureActiveDirectory|Microsoft365\.Audit"
+    r"\.Exchange|Microsoft365\.Audit\.General|Microsoft365\.Audit\.SharePoint|Microsoft365\.DLP"
+    r"\.All|Nginx\.Access|Okta\.SystemLog|OneLogin\.Events|Osquery\.Batch|Osquery\.Differential"
+    r"|Osquery\.Snapshot|Osquery\.Status|OSSEC\.EventInfo|Salesforce\.Login|Salesforce\.LoginAs"
+    r"|Salesforce\.Logout|Salesforce\.URI|Slack\.AccessLogs|Slack\.AuditLogs|Slack"
+    r"\.IntegrationLogs|Sophos\.Central|Suricata\.Anomaly|Suricata\.DNS|Syslog\.RFC3164|Syslog"
+    r"\.RFC5424|Zeek\.DNS)$"
```
lindsey-w, line 50: I think we can add something for custom logtypes too, where they _must_ start with `Custom.`

panther_analysis_tool/main.py
```diff
@@ -815,6 +824,29 @@ def classify_analysis(
...
+                    for each_key in analysis_schema.schema.keys():
+                        if str(each_key) == "Or('LogTypes', 'ScheduledQueries')":
+                            tmp_logtypes_key = each_key
+                            break
+                    tmp_logtypes = analysis_schema.schema[tmp_logtypes_key]
+                    analysis_schema.schema[tmp_logtypes_key] = [str]
+            if ignore_resourcetypes_validation:
+                if "ResourceTypes" in analysis_spec:
+                    tmp_resourcetypes = analysis_schema.schema["ResourceTypes"]
+                    analysis_schema.schema["ResourceTypes"] = [str]
```
lindsey-w, line 840: Is there a specific request/reason we would want to support ignoring this validation? Just wondering if this adds complexity when on the backend it will fail / there isn't a way to skip the validation.

panther_analysis_tool/schemas.py
wey-chiang: Do all custom logtypes start with the `Custom.` prefix?

wey-chiang: That'd remove the need for the `--ignore-logtypes-validation` flag

panther_analysis_tool/main.py
wey-chiang: This was namely added for the case of custom logtypes/schemas

panther_analysis_tool/schemas.py
lindsey-w: I think so: https://docs.runpanther.io/data-onboarding/custom-log-types 

cc @alxarch to confirm?

nhakmiller: They do, we enforce it in a few places but you can't create a custom log type without the prefix `Custom.`

wey-chiang: Wondering if it might make sense to leave this in to preemptively workaround a future divergence between `panther` and `panther_analysis_tool`

nhakmiller: I'd say let's drop the `--ignore-logtypes-validation` and `--ignore-resourcetypes-validation` flags for now. I generally agree we shouldn't force validation that might break workflows, but it seems to add a lot of overhead. We can update `panther_analysis_tool` pretty quickly to address any issues that arise from this.

nhakmiller (approved):
Looks good overall, but I think removing the options to skip validation will help reduce some of the complexity.

PR #97, opened 2021-04-29
Ignore Dot-Files

wey-chiang:
Closes https://app.asana.com/0/1199906407795548/1200249363083484/f

### Background

PAT attempts to load all of the YAML and JSON files found in the specified PATH, this however includes dot-files (nix hidden folders and hidden files)

### Changes

* Added conditional to skip dot-files 

### Testing

* See attached SS (ignore the failing tests, those are non-related 😄 )
 * Created a directory `.bad_rules` which contained a malformed detection `example_bad_rule.yml` and `example_bad_rule.py`
 * Created a malformed detection in the root directory `.example_bad_rule.yml` and `.example_bad_rule.py`

# Testing - Prior to changes

image

# Testing - After changes

image

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -156,6 +156,9 @@ def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any,
     loaded_specs: List[Any] = []
     for directory in directories:
         for relative_path, _, file_list in os.walk(directory):
+            # Skip hidden folders
+            if relative_path.split('/')[-1].startswith("."):
+                continue
             # setup yaml object
             yaml = YAML(typ="safe")
             # If the user runs with no path args, filter to make sure
@@ -179,6 +182,9 @@ def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any,
                     logging.debug("Skipping path %s", relative_path)
                     continue
             for filename in sorted(file_list):
+                # Skip hidden files
+                if filename.startswith("."):
+                    continue
                 spec_filename = os.path.abspath(os.path.join(relative_path, filename))
                 # skip loading files that have already been imported
                 if spec_filename in loaded_specs:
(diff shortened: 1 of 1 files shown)


panther_analysis_tool/main.py
```diff
@@ -156,6 +156,9 @@ def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any,
     loaded_specs: List[Any] = []
     for directory in directories:
         for relative_path, _, file_list in os.walk(directory):
+            # Skip hidden folders
+            if relative_path.split('/')[-1].startswith("."):
+                continue
```
nhakmiller, line 161: Weird edge case, this totally skips testing files in the current working directory (since the path is just `.` which happens to start with `.`). Maybe a slight modification here?
```suggestion
            # Skip hidden folders
            if relative_path.split('/')[-1].startswith(".") and relative_path != '.':
                continue
```

maybe that's a bit hacky, if you have a better suggestion all for it. But I'd like to preserve the functionality that running `panther_analysis_tool test` in a directory with detections tests all those detections.

wey-chiang: That makes sense since the first iteration of `os.walk` will yield `./` so it ignores `.`

The suggestion won't work since `relative_path` will be `./` (since we're not applying the same python splitting to this condition

```suggestion
            # Skip hidden folders
            if relative_path.split('/')[-1].startswith(".") and relative_path != "./":
                continue
```

nhakmiller: Whichever way works is fine with me, I was just testing a few cases locally and noticed it and tried this out to fix it and it seemed work. I'm sure there are some weird edge cases, not sure what the most definitive rule set would be to work perfectly in all cases.

wey-chiang: This should work now. Attached is a SS to show that `./` (the edge case which happens to actually be the first guaranteed response from `os.walk` is the first iteration)

image

If there's a weird issue where this is not guaranteed on all *nix systems, we can always write the conditional as:

```
if relative_path.split('/')[-1].startswith('.') and len(relative_path.split('/')) != 1:
```

To basically assert that there's always something after the initial `./` but I think we should be safe -- the current implementation is small optimization that avoids an extra (likely unnecessary) `len` and `split` call each time

nhakmiller:
For whatever reason, on my system it needs to be compared against `'.'` ¯\_(ツ)_/¯ so I put in both checks.

nhakmiller (approved):

PR #99, opened 2021-05-03
Add requirements

nhakmiller:
### Background

When @georgepsarakis modified the `setup.py` file, it broke new pip installs as it relied on the existence of a file that is not present in the pip package. This PR adds a manifest file which I believe indicates to pip to include this file.

### Changes

* Add manifest file
* Bump to 0.6.1

### Testing

* None, not sure how to test

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include requirements.txt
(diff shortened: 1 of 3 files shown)


lindsey-w (approved):

nhakmiller:
Ok I ran the package commands and just opened up the tarball it created and now it does look like we have the required file 👍

PR #100, opened 2021-05-13
Introduce custom schema uploader

georgepsarakis:
### Background

Relates to https://app.asana.com/0/1200276669536471/1200131236776274

### Changes

* Add an `Uploader` class that manages custom schema updates from YAML files.
  * The `schema` field must be prefixed with `Custom.`
  * If the `referenceURL` and `description` attributes are defined they are used, otherwise the backend values remain unchanged.
* Add a new sub-command for updating / adding custom schemas:

```
usage: panther_analysis_tool update-custom-schemas [-h]
                                                   [--aws-profile AWS_PROFILE]
                                                   [--path PATH]

optional arguments:
  -h, --help            show this help message and exit
  --aws-profile AWS_PROFILE
                        The AWS profile to use when updating the AWS Panther
                        deployment.
  --path PATH           The relative or absolute path to Panther custom
                        schemas.
```

### Testing

* [x] Create two sample YAML files containing schemas and execute the command. The schemas are updated in the Panther UI.
* [x] Malformed YAML errors are reported.
* [x] API errors are reported.


nhakmiller:
I can take a look, but just to double check did you coordinate this with @alxarch? I know he did most of the schema work so far.

georgepsarakis:
> I can take a look, but just to double check did you coordinate this with @alxarch? I know he did most of the schema work so far.

@nhakmiller sure, please take a look and let me know if you agree with the relocation of the fixtures as well. This PR was opened as draft and was a work-in-progress, that's why I hadn't invited Alex to review yet.

Any feedback is welcome!

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+            self._create_lambda_request(
+                endpoint=self._LIST_SCHEMAS_ENDPOINT,
+                payload={
+                    'isManaged': False
+                }
+            )
+        )
+
+    def put_schema(self, name: str, definition: str, revision: int,  # pylint: disable=too-many-arguments
+                   description: str, reference_url: str) -> Tuple[bool, dict]:
```
kostaspap, line 43: Having the revision be part of the request makes things more complicated dealing with this API. Not part of this PR but logtypes-api API codebase would be really simplified if we omitted that requirement.

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@dataclass
+class UploaderResult:
+    filename: str
+    name: Optional[str]
+    api_response: Optional[Dict[str, Any]] = None
+    definition: Optional[Dict[str, Any]] = None
+    error: Optional[str] = None
+    existed: Optional[bool] = None
```
kostaspap, line 83: It would be useful to add some small comments what each parameter contains, for future reference.

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+            results.append(result)
+        return results
+
+    @staticmethod
+    def _load_from_yaml(files: List[str]) -> Dict[str, ProcessedFile]:
+        yaml_parser = YAML(typ="safe")
+
+        processed_files = {}
+        for filename in files:
+            logger.info('Processing schema in file %s', filename)
```
kostaspap, line 172:
```suggestion
            logger.info('Loading schema from file %s', filename)
```

Maybe that output makes more sense from a user's perspective?

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+                )
+
+        for filename, processed_file in processed_files.items():
+            if processed_file.error is not None:
+                continue
+
+            name, error = self._extract_schema_name(processed_file.yaml)
+            result = UploaderResult(filename=filename, name=name)
+
+            result.name = name
```
kostaspap, line 152: this is already set?

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+            result.error = error
```
kostaspap, line 153: You could move that to the one liner, in line 150?

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+            if not result.error:
+                existed, success, response = self._update_or_create_schema(name, processed_file)
+                result.existed = existed
+                if not success:
+                    api_error = response.get('error')
+                    if api_error is not None:
```
kostaspap, line 159: What if it is `None`? Should we set the `result.error` to some default value?

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+            name=name,
+            definition=processed_file.raw,
+            revision=current_revision,
+            reference_url=reference_url,
+            description=description,
+        )
+        return existed, success, response
+
+
+def discover_files(base_path: str, pattern: str = '*.yml') -> List[str]:
```
kostaspap, line 237: I would also say include `*.yaml` in there. We support `*.yaml` in other tools (e.g. `pantherlog`)

tests/fixtures/custom-schemas/invalid/schema-1.yml
```diff
@@ -0,0 +1,14 @@
+schema: Custom.SampleSchema1
```
kostaspap, line 1: I fail to see.. why is this invalid?

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -0,0 +1,279 @@
...
+            self._existing_schemas = response['results']
+        return self._existing_schemas
+
+    def find_schema(self, name: str) -> Optional[Dict[str, Any]]:
+        for schema in self.existing_schemas:
+            if schema['name'] == name:
+                return schema
+        return None
+
+    def process(self) -> List[UploaderResult]:
```
kostaspap, line 129: Could you add description to this process? I am particularly concerned about the logic that first fetches schemas before uploading them. It will not be obvious to everyone why this has to happen.

kostaspap (commented):
Nice work! I really like the thorough tests. 
Could you also paste the output when running this tool in the following scenario: 
1. You upload valid schema(s)
2. You try to upload invalid schemas.

panther_analysis_tool/log_schemas/user_defined.py
georgepsarakis: Sure, I will write some notes on the matching revision requirement on behalf of the API endpoint.

tests/fixtures/custom-schemas/invalid/schema-1.yml
georgepsarakis: It's not actually, I will add a comment to clarify. The point was to mix valid and invalid schemas together and test the behavior.

panther_analysis_tool/log_schemas/user_defined.py
georgepsarakis: Sure, will do.

panther_analysis_tool/log_schemas/user_defined.py
georgepsarakis: We enter this path only if `result.error` is not present, so we don't need to reset it, but only keep record of the API error if there's one.

panther_analysis_tool/log_schemas/user_defined.py
georgepsarakis: I think it was introduced to address some concurrency concerns, no? As a side note, there is also a great validation for detecting backwards incompatibilities, but I think it is based on the assumption that updates are performed sequentially (because it needs to compare with the previous version naturally). For such extra-low throughput / rarely called endpoints that perform critical updates, perhaps a pessimistic lock would be a good approach.

alxarch (commented):
I am concerned with having the tool bulk-upload schemas.
I would prefer for this tool to operate on a per-schema basis.

Each schema update triggers a datacatalog update for each updated schema.
These updates are queued and there is no way to know if the sync has completed.
Having multiple such updates running in parallel, especially if it involves schemas that have accumulated many partitions is not a good idea for the stability of the system.
While we have done our best to ensure updates are idempotent, there are destructive operations that can happen in parallel during the update such as the re-generation of views.

Fixing this BE side would be extremely time consuming and we have decided to not allocate resources on Glue/Athena quirks.

For bulk upload to be safe we would need to be sure that at least the UpdateTable request has been processed by the data catalog updater. 
To properly solve this the datacatalog updater would need to persist the state of ongoing update tasks and have the client poll for completed status before proceeding with the next update.

One hacky way I can think would involve stamping the table metadata with the schema revision of the update.
The client would have to poll the Glue data catalog  (or a different logtypes API endpoint) until the table is on a revision >= the update revision.

I understand such a thing would be an overkill for this tool.
Let's cross our fingers and hope for the best. :crossed_fingers:

alxarch (approved):
The race issues I mentioned should be solved elsewhere.
A delay as suggested by me and corrected by @kostaspap would not help due to the batching delay in the datacatalog queue.
Here's a stamp! :1st_place_medal: 

On a side note:
Turns out managed schema updates do not trigger a data catalog update for the updated schemas :scream: . I have my work cut out for me it seems!

nhakmiller (approved):
I have no issues with these changes, and since @alxarch approved I will approve as well to allow the merge.

PR #101, opened 2021-06-01
Use Rule Engine core structures when executing tests

georgepsarakis:
### Background

Closes https://app.asana.com/0/1200360676535738/1200267145879882 

### Changes

* Transfer `DataModel` code from Rule Engine and its tests.
* Transfer immutable / case-insensitive data structures and their tests.
* Transfer the event wrapper `PantherEvent`, containing the Data Model access method `udm`.
* Remove previous substitute implementations for the previous structures.
* Integrate changes in `main` functions.

### Testing

* Executed tests on `panther-analysis` with [this fix](https://github.com/panther-labs/panther-analysis/pull/273) applied.
* Unit tests


tests/fixtures/valid_analysis/data_models/example_data_model_python.yml
```diff
@@ -1,7 +1,7 @@
 AnalysisType: datamodel
 DataModelID: GSuite.Events.DataModel
 Enabled: true
-Filename: example_data_model_python.py
+Filename: GSuite.Events.DataModel.py
```
lindsey-w, line 4: curious, why the name change?

lindsey-w (approved):
Lots of good stuff here

tests/fixtures/valid_analysis/data_models/example_data_model_python.yml
nhakmiller: I agree, unless there is a specific reason I'd prefer to keep extraneous `.`'s out of filenames.

nhakmiller (approved):
Wow tons of new testing and killing our monolith main file! You love to see it 🚀

wey-chiang:
Well this looks super promising! Nice work 😄 🚀

tests/fixtures/valid_analysis/data_models/example_data_model_python.yml
wey-chiang: +1 bump

wey-chiang (approved):
:D

panther_analysis_tool/data_model.py
```diff
@@ -0,0 +1,125 @@
...
+E_NO_DATA_MODEL_FOUND = "a data model hasn't been specified for log type"
+
+# constants used to extract data from data model
+NAME = 'name'
+PATH = 'path'
+METHOD = 'method'
+
+
+# Temporary alias for compatibility
+get_logger = logging.getLogger
```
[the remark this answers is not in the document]
georgepsarakis: Logging configuration is handled in [`main`](https://github.com/panther-labs/panther_analysis_tool/blob/e11228f7d2c1115a3cfde6fec01c7cb6a4a767be/panther_analysis_tool/main.py#L1396). Note that `getLogger` is a function and is called when the `DataModel` is created, so it will acquire the centralized logging configuration correctly.

Or is it something else that you were concerned about maybe?

tests/fixtures/valid_analysis/data_models/example_data_model_python.yml
georgepsarakis: Unfortunately there is an important reason for renaming. The `DataModel` constructor fails if it cannot create a module and uses a specific naming convention for the temporary file it stores. The reason for the failure is that the fake filesystem does not patch the `importlib` loaders, because these functions actually operate on a VM level. I assume it would actually be insecure if that was allowed as someone could shadow a known package dynamically. Thus, we need to rename the file to actually exist in the filesystem as the `DataModel` expects.

I expect that it may need some a few more tweaks until we can fully align both codebases, so we could fix this in a subsequent PR, no problem with that. 

Note that it would be an improvement if the constructor performed no operations, but instead just initialized attributes. Why? Because if the constructor calls instance functions, especially those with side-effects or external dependencies (e.g. filesystem), that could fail, it becomes inherently difficult to test, because you cannot make reliable assertions on an object that cannot be constructed. The pattern that is followed in such cases is to use a property getter and check if a protected/private attribute is already populated:

```python
@property
def module(self):
   if self._module is None:
     self._module = self._get_module()
   return self._module
```

PR #103, opened 2021-06-07
PAT NotEQuals Filtering & Skip Test Disabled Options

wey-chiang:
### References

Closes https://app.asana.com/0/1200360676535738/1200356830724953/f

### Background

Request for the ability to use `!=` for filtering tags as well as an extra optional flag to skip tests for disabled detections.

### Changes

* Added inverted filtering functionality
* Added flag that allows for the skipping of tests for disabled detections

### Testing

* `make test`
* `make integration`

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -279,7 +279,7 @@ def zip_analysis(args: argparse.Namespace) -> Tuple[int, str]:
                 analysis.append((file_name, f_path, spec))
                 files.add(file_name)
                 files.add("./" + file_name)
-        analysis = filter_analysis(analysis, args.filter)
+        analysis = filter_analysis(analysis, args.filter, args.filter_inverted)
         for analysis_spec_filename, dir_name, analysis_spec in analysis:
             zip_out.write(analysis_spec_filename)
             # datamodels may not have python body
@@ -669,11 +669,13 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
         return 1, ["Nothing to test in {}".format(args.path)]
 
     # Apply the filters as needed
+    if getattr(args, "filter_inverted", None) is None:
+        args.filter_inverted = {}
     for key in specs:
-        specs[key] = filter_analysis(specs[key], args.filter)
+        specs[key] = filter_analysis(specs[key], args.filter, args.filter_inverted)
 
     if all((len(specs[key]) == 0 for key in specs)):
-        return 1, ["No analysis in {} matched filters {}".format(args.path, args.filter)]
+        return 1, ["No analysis in {} matched filters {} - {}".format(args.path, args.filter, args.filter_inverted)]
 
     # import each data model, global, policy, or rule and run its tests
     # first import the globals
@@ -829,7 +831,7 @@ def print_summary(
             print(err_message.format(spec_filename, spec_error))
 
 
-def filter_analysis(analysis: List[Any], filters: Dict[str, List]) -> List[Any]:
+def filter_analysis(analysis: List[Any], filters: Dict[str, List], filters_inverted: Dict[str, List]) -> List[Any]:
     if filters is None:
         return analysis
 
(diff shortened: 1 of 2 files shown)


wey-chiang:
Here's some Screenshots from testing

# Before neq Filtering

image

# After neq Filtering

image

# Before Skip Disabled Tests

image

# After Skip Disabled Tests

image

panther_analysis_tool/main.py
```diff
@@ -669,11 +669,15 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
         return 1, ["Nothing to test in {}".format(args.path)]
 
     # Apply the filters as needed
+    if getattr(args, "filter_inverted", None) is None:
+        args.filter_inverted = {}
```
nhakmiller, line 673: Why do you do this input normalization here and not in the `zip_analysis` function when you use this argument?

If it's because it only needs to happen once and test always happens before zip, I think we should we move this setup to the `def run()` function on line ~1500.

nhakmiller (approved):
LGTM! One small question. Thanks for jumping on this so quick.

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -258,7 +258,17 @@ def test_validate_outputs(self):
     def test_with_tag_filters(self):
         args = pat.setup_parser().parse_args(
             f'test --path {DETECTIONS_FIXTURES_PATH}/valid_analysis --filter Tags=AWS,CIS'.split())
-        args.filter = pat.parse_filter(args.filter)
+        args.filter, args.filter_inverted = pat.parse_filter(args.filter)
+        return_code, invalid_specs = pat.test_analysis(args)
+        assert_equal(return_code, 0)
+        assert_equal(len(invalid_specs), 0)
+
+    def test_with_tag_filters_inverted(self):
```
nhakmiller, line 266: 👍

panther_analysis_tool/main.py
```diff
@@ -757,11 +761,14 @@ def setup_data_models(data_models: List[Any]) -> Tuple[Dict[str, DataModel], Lis
...
 
 def setup_run_tests(
-    log_type_to_data_model: Dict[str, DataModel], analysis: List[Any], minimum_tests: int
+    log_type_to_data_model: Dict[str, DataModel], analysis: List[Any], minimum_tests: int,
+        skip_disabled_tests: bool
 ) -> Tuple[DefaultDict[str, List[Any]], List[Any]]:
     invalid_specs = []
     failed_tests: DefaultDict[str, list] = defaultdict(list)
     for analysis_spec_filename, dir_name, analysis_spec in analysis:
+        if skip_disabled_tests and not analysis_spec.get("Enabled", False):
```
lindsey-w, line 770: does this mean by default we will skip the test? Should it be the opposite? I suppose it should never happen since enabled is required anyway 🤷🏻‍♀️

lindsey-w (approved):
Really useful features!

panther_analysis_tool/main.py
wey-chiang: `skip_disabled_tests` will default to `False` when the flag is not set. This will only skip tests when that flag is set and the detection is `Enabled` is `False`

panther_analysis_tool/main.py
wey-chiang: The reason I did not put this in the `run` function is because the unit tests directly invoke `zip_analysis`

I'll move it into `zip_analysis` to avoid adding extra complexity to the tests

panther_analysis_tool/main.py
lindsey-w: ahh yea, I was thinking when `skip_disable_tests` is set, but an analysis spec doesn't specify `Enabled` status: seems like this, by default, will skip the test. Since `enabled` is a required field - shouldn't really matter.  Just an observation/question about understanding the logic I suppose.

panther_analysis_tool/main.py
wey-chiang: Moved this down into `run` and adjusted the failing unit tests to manually set the `args.filter_inverted`

panther_analysis_tool/main.py
wey-chiang: Ahh gotcha, yeah definitely a good call out! I feel more comfortable with this implementation knowing that `enabled` is a required field

PR #109, opened 2021-06-24
Port Panther Rules Engine rule execution logic and flow

georgepsarakis:
Closes https://app.asana.com/0/1200441743569436/1200267145879888

### Background

In order to mirror the behavior during of the core rule executor and eventually share the code between Panther Engine and PAT, the corresponding module is being transferred from the Engine here. The previous rule execution logic must be adapted / replaced by encapsulated functionality in the `Rule` class.

### Changes

* Add an integration test that checks that all detections defined `panther-analysis` pass.
* Allow a path to be passed directly to the `Rule` executor for importing the rule code as module, rather than creating a new temporary file containing the raw code.
* Adapt test setup procedure to instantiate `Rule` and utilize the `RuleResult` for evaluating and displaying execution results. 

### Testing

* Unit tests
* Run tests on `panther-analysis`: `panther_analysis_tool test --path .`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,11 @@ unit:
 
 integration:
 	pipenv run panther_analysis_tool test --path tests/fixtures/detections/valid_analysis
+	rm -rf panther-analysis
+	git clone https://github.com/panther-labs/panther-analysis.git
+	cd panther-analysis && cat requirements.txt | grep -v 'panther-analysis-tool==' > requirements.ci.txt
+	cd panther-analysis && pip install -r requirements.ci.txt
+	cd panther-analysis && pipenv run panther_analysis_tool --version && pipenv run panther_analysis_tool test --path .
 
 test: unit
 
--- a/panther_analysis_tool/destination.py
+++ b/panther_analysis_tool/destination.py
@@ -1,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class FakeDestination:
+    """Stub class as a replacement for the Destination class
+    that wraps alert output metadata."""
+
+    destination_id: str
+    destination_display_name: str
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -124,17 +124,6 @@ SET_FIELDS = [
     "Tags",
 ]
 
-# Available destination names for use by rules in tests
-AVAILABLE_DESTINATION_NAMES = ("ExampleDestinationName",)
-
-# The rule executor needs to map any destination names returned by the `destination` function
-# to actual destination objects. We are using a fake Destination dataclass, since we
-# just need the name and identifier fields.
-OUTPUT_DISPLAY_NAMES = {
-    name: FakeDestination(destination_id=str(uuid4()), destination_display_name=name)
-    for name in AVAILABLE_DESTINATION_NAMES
-}
-
 
 # exception for conflicting ids
 class AnalysisIDConflictException(Exception):
@@ -721,6 +710,15 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
             f"No analysis in {args.path} matched filters {args.filter} - {args.filter_inverted}"
         ]
 
+    available_destinations = []
+    if args.available_destination:
+        available_destinations.extend(args.available_destination)
+
+    destinations_by_name = {
+        name: FakeDestination(destination_id=str(uuid4()), destination_display_name=name)
+        for name in available_destinations
+    }
+
     # import each data model, global, policy, or rule and run its tests
     # first import the globals
     #   add them sys.modules to be used by rule and/or policies tests
@@ -733,7 +731,8 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
 
     # then, import rules and policies; run tests
     failed_tests, invalid_detection = setup_run_tests(
-        log_type_to_data_model, specs[DETECTION], args.minimum_tests, args.skip_disabled_tests
+        log_type_to_data_model, specs[DETECTION], args.minimum_tests, args.skip_disabled_tests,
+        destinations_by_name=destinations_by_name
     )
     invalid_specs.extend(invalid_detection)
 
@@ -809,6 +808,7 @@ def setup_run_tests(
     analysis: List[Any],
     minimum_tests: int,
     skip_disabled_tests: bool,
+    destinations_by_name: Dict[str, FakeDestination]
 ) -> Tuple[DefaultDict[str, List[Any]], List[Any]]:
     invalid_specs = []
     failed_tests: DefaultDict[str, list] = defaultdict(list)
@@ -839,7 +839,7 @@ def setup_run_tests(
                 )
             )
             analysis_funcs["run"] = functools.partial(
-                rule.run, outputs={}, outputs_names=OUTPUT_DISPLAY_NAMES, batch_mode=False
+                rule.run, outputs={}, outputs_names=destinations_by_name, batch_mode=False
             )
             analysis_funcs["module"] = rule.module
 
@@ -849,6 +849,7 @@ def setup_run_tests(
             log_type_to_data_model,
             failed_tests,
             minimum_tests,
+            bool(destinations_by_name)
         )
         print("")
     return failed_tests, invalid_specs
@@ -1062,6 +1063,7 @@ def run_tests(
     analysis_data_models: Dict[str, DataModel],
     failed_tests: DefaultDict[str, list],
     minimum_tests: int,
+    destination_names_strict_check: bool,
 ) -> DefaultDict[str, list]:
 
     if len(analysis.get("Tests", [])) < minimum_tests:
@@ -1077,7 +1079,11 @@ def run_tests(
         print("\tNo tests configured for {}".format(analysis_id))
         return failed_tests
 
-    failed_tests = _run_tests(analysis, analysis_funcs, analysis_data_models, failed_tests)
+    failed_tests = _run_tests(analysis,
+                              analysis_funcs,
+                              analysis_data_models,
+                              failed_tests,
+                              destination_names_strict_check)
 
     if minimum_tests > 1 and not (
         [x for x in analysis["Tests"] if x["ExpectedResult"]]
@@ -1095,6 +1101,7 @@ def _run_tests(
     analysis_funcs: Dict[str, Any],
     analysis_data_models: Dict[str, DataModel],
     failed_tests: DefaultDict[str, list],
+    destination_names_strict_check: bool
 ) -> DefaultDict[str, list]:
     is_policy = analysis.get("PolicyID") is not None
     analysis_id = analysis.get("PolicyID") or analysis["RuleID"]
@@ -1149,107 +1156,30 @@ def _run_tests(
                 test_result["outcome"] = "FAIL"
                 failed_tests[analysis_id].append(unit_test["Name"])
 
+            # validate reserved function return types and values
+            # Only applies to rules which match an incoming event
+            if unit_test["ExpectedResult"]:
+                for function_name in RESERVED_FUNCTIONS:
+                    function_error = getattr(rule_result, f"{function_name}_exception")
+                    # For backwards compatibility we can accept invalid destination names,
+                    # as long as a string is returned
+                    if function_name == "destinations" and function_error is not None:
+                        # Strict check is enabled when users pass destination names
+                        # through command-line parameters
+                        if not destination_names_strict_check:
+                            if isinstance(function_error, ValueError):
+                                message = function_error.args[0]
+                                if re.match(r"Invalid Destinations: \['.+'\]", message):
+                                    function_error = None
+
+                    if function_error is not None:
+                        test_result[function_name] = test_result["outcome"] = "FAIL"
+                        failed_tests[analysis_id].append(f"{unit_test['Name']}:{function_name}")
+
         # print results
         print("\t[{}] {}".format(test_result["outcome"], unit_test["Name"]))
 
-        # validate reserved function return types and values
-        # Only applies to rules which match an incoming event
-        if unit_test["ExpectedResult"] and not is_policy:
-            for function_name in RESERVED_FUNCTIONS:
-                if getattr(rule_result, f"{function_name}_exception") is not None:
-                    test_result[function_name] = test_result["outcome"] = "FAIL"
-                    failed_tests[analysis_id].append(f"{unit_test['Name']}:{function_name}")
     return failed_tests
-
-
-def _verify_test_run(
-    event: PantherEvent,
-    mock_methods: Dict[str, MagicMock],
-    analysis_funcs: Dict[str, Any],
-) -> Dict[str, Any]:
-    test_run_result: Dict[str, str] = defaultdict(lambda: "PASS")
-    for func_name, func in analysis_funcs.items():
-        if func_name not in RESERVED_FUNCTIONS:
-            continue
-        try:
-            if mock_methods:
-                with patch.multiple(analysis_funcs["module"], **mock_methods):
-                    func_out = func(event)
-                    if not func_out:
-                        test_run_result[func_name] = "FAIL"
-            else:
-                func_out = func(event)
-                if not func_out:
-                    test_run_result[func_name] = "FAIL"
-        except Exception as err:  # pylint: disable=broad-except
-            func_out = err
-            test_run_result[func_name] = "FAIL"
-
-        func_out_valid_type = True
-        valid_output, func_out = validate_outputs(func_name, func_out)
-        if not valid_output or isinstance(func_out, (AttributeError, AssertionError)):
-            func_out_valid_type = False
-            test_run_result[func_name] = "FAIL"
-
-        print(
-            f"\t\t[{test_run_result[func_name]}] "
-            f"{'' if func_out_valid_type else '[INVALID TYPE] '}"
-            f"[{func_name}] {func_out}"
-        )
-    return test_run_result
-
-
-def validate_outputs(function_name: str, function_output: Any) -> Tuple[bool, Any]:
-    # Defaults to valid and function output
-    # Invalidating criteria will overwrite these values
-    is_valid = True
-    output = function_output
-    if function_name == "severity":
-        # Type checking for severity
-        if not isinstance(function_output, str):
-            is_valid = False
-        # Value Validation for severity
-        if str(function_output).upper() not in VALID_SEVERITIES:
-            is_valid = False
-            output = AssertionError(
-                f"Expected [{function_name}] to be any of the following: "
-                f"{str(VALID_SEVERITIES)}, got [{str(function_output)}] instead."
-            )
-    elif function_name == "alert_context":
-        is_valid = isinstance(function_output, dict)
-    elif function_name == "destinations":
-        # Type checking for destinations
-        if isinstance(function_output, list):
-            # Type checking for elements in list
-            # Note: we cannot perform value validation unless we were to query the outputs-api
-            if not all(isinstance(x, str) for x in function_output):
-                # List contains non-string element, so enumerate the types present for error msg
-                list_types = set()
-                for each_item in function_output:
-                    list_types.add(type(each_item))
-                is_valid = False
-                output = AttributeError(
-                    f"Expected [{function_name}] to return a list of strings, "
-                    f"got types [{list_types}] instead"
-                )
-        else:
-            is_valid = False
-    # For reserved functions besides severity and destinations, we expect a string output
-    else:
-        is_valid = isinstance(function_output, str)
-
-    # Invalid Types, used to generate a corresponding error message
-    if not is_valid and output == function_output:
-        if function_name == "destinations":
-            expected_type = "list"
-        else:
-            expected_type = "string"
-
-        output = AttributeError(
-            f"Expected [{function_name}] to return a [{expected_type}], "
-            f"got [{type(function_output)}] instead"
-        )
-    return is_valid, output
 
 
 def setup_parser() -> argparse.ArgumentParser:
@@ -1313,6 +1243,15 @@ def setup_parser() -> argparse.ArgumentParser:
         "type": strtobool,
         "help": "Meant for advanced users; allows skipping of extra keys from schema validation.",
     }
+    available_destination_name = '--available-destination'
+    available_destination_arg: Dict[str, Any] = {
+        "required": False,
+        "default": None,
+        "type": str,
+        "action": "append",
+        "help": "A destination name that may be returned by the destination() function. "
+                "Repeat the argument to define more than one name.",
+    }
 
     parser = argparse.ArgumentParser(
         description="Panther Analysis Tool: A command line tool for "
@@ -1337,6 +1276,7 @@ def setup_parser() -> argparse.ArgumentParser:
     release_parser.add_argument(path_name, **path_arg)
     release_parser.add_argument(skip_test_name, **skip_test_arg)
     release_parser.add_argument(skip_disabled_test_name, **skip_disabled_test_arg)
+    release_parser.add_argument(available_destination_name, **available_destination_arg)
     release_parser.set_defaults(func=generate_release_assets)
 
     test_parser = subparsers.add_parser(
@@ -1347,6 +1287,7 @@ def setup_parser() -> argparse.ArgumentParser:
     test_parser.add_argument(path_name, **path_arg)
     test_parser.add_argument(ignore_extra_keys_name, **ignore_extra_keys_arg)
     test_parser.add_argument(skip_disabled_test_name, **skip_disabled_test_arg)
+    test_parser.add_argument(available_destination_name, **available_destination_arg)
     test_parser.set_defaults(func=test_analysis)
 
     publish_parser = subparsers.add_parser(
@@ -1392,6 +1333,7 @@ def setup_parser() -> argparse.ArgumentParser:
     publish_parser.add_argument(out_name, **out_arg)
     publish_parser.add_argument(skip_test_name, **skip_test_arg)
     publish_parser.add_argument(skip_disabled_test_name, **skip_disabled_test_arg)
+    publish_parser.add_argument(available_destination_name, **available_destination_arg)
     publish_parser.set_defaults(func=publish_release)
 
     upload_parser = subparsers.add_parser(
@@ -1405,6 +1347,7 @@ def setup_parser() -> argparse.ArgumentParser:
     upload_parser.add_argument(skip_test_name, **skip_test_arg)
     upload_parser.add_argument(skip_disabled_test_name, **skip_disabled_test_arg)
     upload_parser.add_argument(ignore_extra_keys_name, **ignore_extra_keys_arg)
+    upload_parser.add_argument(available_destination_name, **available_destination_arg)
     upload_parser.set_defaults(func=upload_analysis)
 
     update_managed_schemas_parser = subparsers.add_parser(
@@ -1422,6 +1365,7 @@ def setup_parser() -> argparse.ArgumentParser:
     zip_parser.add_argument(path_name, **path_arg)
     zip_parser.add_argument(skip_test_name, **skip_test_arg)
     zip_parser.add_argument(skip_disabled_test_name, **skip_disabled_test_arg)
+    zip_parser.add_argument(available_destination_name, **available_destination_arg)
     zip_parser.set_defaults(func=zip_analysis)
 
     zip_schemas_parser = subparsers.add_parser(
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -19,12 +19,14 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import argparse
 import base64
+import functools
 import hashlib
 import importlib.util
 import json
 import logging
 import mimetypes
 import os
+import pathlib
 import re
 import subprocess  # nosec
 import sys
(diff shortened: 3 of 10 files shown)


panther_analysis_tool/rule.py
```diff
@@ -0,0 +1,728 @@
...
+        path = id_to_path(tmp_dir, identifier)
+        store_modules(path, body)
+        return self.from_file(identifier, path)
+
+    @abstractmethod
+    def get_module(self, identifier: str, resource: str) -> ModuleType:
+        pass
+
+
+class FilesystemImporter(BaseImporter):
```
georgepsarakis, line 180: Test coverage required for these classes.

panther_analysis_tool/rule.py
```diff
@@ -0,0 +1,728 @@
...
+
+        if (
+            not "analysisType" in config
+        ):  # backwards compatible check for data from before we returned this
+            self.rule_type = TYPE_RULE
+        else:
+            self.rule_type = config["analysisType"]
+
+        self.rule_id = config["id"]
+        # self.rule_body = config['body']
```
georgepsarakis, line 240: ✂️

panther_analysis_tool/destination.py
```diff
@@ -0,0 +1,7 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class FakeDestination:
```
georgepsarakis, line 5: Add comments.

georgepsarakis:
Ready for review.

panther_analysis_tool/main.py
```diff
@@ -807,21 +821,18 @@ def setup_run_tests(
...
+                dict(
+                    id=analysis_id,
+                    analysisType=analysis_type,
+                    path=pathlib.Path(module_code_path),
+                    versionId="0000-0000-0000",
+                )
+            )
+            analysis_funcs["run"] = functools.partial(
+                rule.run, outputs={}, outputs_names=OUTPUT_DISPLAY_NAMES, batch_mode=False
+            )
```
lindsey-w, line 834: This is neat, I hadn't seen `functool.partial()` before.

panther_analysis_tool/main.py
```diff
@@ -118,6 +123,13 @@
     "Tags",
 ]
 
+AVAILABLE_DESTINATION_NAMES = ("ExampleDestinationName",)
+
+OUTPUT_DISPLAY_NAMES = {
+    name: FakeDestination(destination_id=str(uuid4()), destination_display_name=name)
+    for name in AVAILABLE_DESTINATION_NAMES
+}
```
lindsey-w, line 131: Might be nice to add a comment describing why we are using these. It took me a few minutes to piece together why this was needed (rule.run(), destination dataclass, this...), but maybe that's just me... 🤷🏻‍♀️

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -35,6 +36,9 @@
 
 print('Using fixtures path:', FIXTURES_PATH)
 
+if os.environ.get('PANTHER_ANALYSIS_TOOL_DEBUG'):
```
lindsey-w, line 39: Is this a env variable you are setting manually? Is the idea to set this env variable _or_ use the `--debug` command line option?

lindsey-w (commented):
LGTM 🥳 

Did you run `make fmt`? I see a few single quotes that might be turned into double quotes with the formatter.  I didn't take as close a look at the `rule` and `test_rule` files - those were just copied over, right?

georgepsarakis:
> Did you run make fmt? I see a few single quotes that might be turned into double quotes with the formatter.

I did, but I'll check again! Maybe we should make this a CI check at some point? Run it and fail if `git diff` finds any differences ?

>  I didn't take as close a look at the rule and test_rule files - those were just copied over, right?

For the most part, yes they are the same [except for the rule module import logic](https://github.com/panther-labs/panther_analysis_tool/pull/109/files#diff-acce689b4ddfb057c8806c0ce5a84fcf19b7163d8dee29fd1d4155d50716a415R291), which can now work by providing a path and not only raw source code (`body`). This is required due to `pyfakefs` unfortunately, but it may be valid for the Rule Engine as well in the future, if we ever switch to having the modules available in the filesystem instead.

panther_analysis_tool/main.py
georgepsarakis: Yes, you are right, I'll explain it here briefly.

tests/unit/panther_analysis_tool/test_main.py
georgepsarakis: Well, that's only for the tests, but I think I'll remove it altogether and maybe revisit in a different PR. It is generally useful to activate debug-level logging, in case something breaks, as it is sometimes difficult to find the root cause for the high-level test failures.

georgepsarakis:
@lindsey-w I ran `make fmt`, no changes required.

panther_analysis_tool/main.py
```diff
@@ -118,6 +123,17 @@
...
+# Available destination names for use by rules in tests
+AVAILABLE_DESTINATION_NAMES = ("ExampleDestinationName",)
+
+# The rule executor needs to map any destination names returned by the `destination` function
+# to actual destination objects. We are using a fake Destination dataclass, since we
+# just need the name and identifier fields.
+OUTPUT_DISPLAY_NAMES = {
+    name: FakeDestination(destination_id=str(uuid4()), destination_display_name=name)
+    for name in AVAILABLE_DESTINATION_NAMES
+}
```
nhakmiller, line 135: So in all cases, it appears the only destination that panther detection unit tests can return is `ExampleDestinationName`?

In general, detection unit tests should be able to return any string since we don't know what users may have named their detections. Is their a way to relax this requirement? Or at least we should have some configurability for end users on this list of destination names.

georgepsarakis: That's a very good observation. Maybe we should add a command-line parameter for now to allow users to define a list of available names. Or perhaps ignore destinations altogether for the time being? I remember you mentioned before that due to their dynamic nature, it might be a bit difficult to handle them at the moment.

nhakmiller: In my opinion the ideal behavior for now is to just verify that a string is returned, which is the current behavior (prior to this change).

If that's not possible, ignoring it might be best. The problem is that if we put this in, even with giving users the option to configure it, it will fail CI/CD build jobs for any users that are using the custom destinations feature until they configure it.

If it's not too much work, can we add the option to configure a list of destination names (either as a command line flag or in a configuration file like `destinations.txt` or something), and if none are configured then skip the destination checks altogether? That way we can tell users to start migrating towards defining destinations, without a hard cutover that will immediately break CI pipelines (since most customers pull the latest version of `panther_analysis_tool` in their pipelines).

georgepsarakis: @nhakmiller I added an `--available-destination <NAME>` parameter that can be repeated in order to provide for multiple values. I removed the default, constant destination name and instead tried to match that `destinations` returns a list of strings by inspecting the corresponding exception message. The check is far from perfect, but I thought it could suffice for now. I didn't want to modify the rule logic at the moment, but if we have to we can do that too.

Note that only when a user provides an `--available-destination` parameter, is strict checking for destination names activated, in which case they'll start to get failures if names do not match. 

Let me know if this approach could work.

nhakmiller: Yeah that's perfect, I'll re-review & probably approve today.

Makefile
```diff
@@ -33,6 +33,11 @@ unit:
 
 integration:
 	pipenv run panther_analysis_tool test --path tests/fixtures/detections/valid_analysis
+	rm -rf panther-analysis
+	git clone https://github.com/panther-labs/panther-analysis.git
+	cd panther-analysis && cat requirements.txt | grep -v 'panther-analysis-tool==' > requirements.ci.txt
+	cd panther-analysis && pip install -r requirements.ci.txt
+	cd panther-analysis && pipenv run panther_analysis_tool --version && pipenv run panther_analysis_tool test --path .
```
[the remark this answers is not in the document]
wey-chiang: Nice!  🙌

panther_analysis_tool/destination.py
```diff
@@ -0,0 +1,10 @@
...
+    """Stub class as a replacement for the Destination class
+    that wraps alert output metadata."""
+
+    destination_id: str
+    destination_display_name: str
```
wey-chiang, line 10: Perhaps we could make some changes to the BE `outputs-api` to allow anyone to get a list of the destination display names, that way we could at least check to see if a destination display name matches some entry in the API.

Thoughts? @nhakmiller  @lindsey-w

panther_analysis_tool/main.py
```diff
@@ -700,6 +705,15 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
...
 
+    available_destinations = []
+    if args.available_destination:
+        available_destinations.extend(args.available_destination)
+
+    destinations_by_name = {
+        name: FakeDestination(destination_id=str(uuid4()), destination_display_name=name)
+        for name in available_destinations
+    }
+
```
wey-chiang, line 716: Ideally we shouldn't have to do this, but it's a nice workaround in the mean time

panther_analysis_tool/main.py
```diff
@@ -1394,6 +1374,7 @@ def setup_parser() -> argparse.ArgumentParser:
     zip_parser.add_argument(path_name, **path_arg)
     zip_parser.add_argument(skip_test_name, **skip_test_arg)
     zip_parser.add_argument(skip_disabled_test_name, **skip_disabled_test_arg)
+    zip_parser.add_argument(available_destination_name, **available_destination_arg)
```
wey-chiang, line 1377: Shouldn't this flag also be added to `upload_parser`?

tests/unit/panther_analysis_tool/test_rule.py
```diff
@@ -0,0 +1,694 @@
...
+from unittest import TestCase
+import tempfile
+from types import ModuleType
+
+from panther_analysis_tool.rule import MAX_DEDUP_STRING_SIZE, MAX_GENERATED_FIELD_SIZE, \
+    Rule, RuleResult, TRUNCATED_STRING_SUFFIX, FilesystemImporter, RawStringImporter
+from panther_analysis_tool.enriched_event import PantherEvent
+
+
+class TestRule(TestCase):  # pylint: disable=too-many-public-methods
```
wey-chiang, line 24: Awesome! Now all we need is some way to synchronize changes from the rules-engine to this

wey-chiang (dismissed):
Nice job! Looks good for the most part, deferring approval to another member on sec-eng

panther_analysis_tool/main.py
georgepsarakis: Yes, it should, maybe we should add a test for the `upload` sub-command using mocking.

tests/unit/panther_analysis_tool/test_rule.py
georgepsarakis: It is the other way around 😄 , the Rules Engine depends on these modules now. You can make a change here and it will be available in the Rules Engine too.

panther_analysis_tool/destination.py
georgepsarakis: I assume that destinations rarely change, right? Perhaps only an export of the id/name in YAML could suffice.

panther_analysis_tool/main.py
georgepsarakis: Can you explain your concern? It's not only a workaround, it also facilitates more reliable testing of the `destinations` functions output, isn't that right? Maybe I am missing something, but you could also consider it an improvement 😄 .

georgepsarakis:
It seems that I accidentally removed the display of the output for rule auxiliary functions. It has been restored in 8121d214f8a0993493247211d1dbdae59359a8dc. Given the extent of changes, I'll open a new PR that simply copies `rule.py` and `test_rule.py`, so that we can review the changes here with more focus and precision after a rebase.

georgepsarakis:
Opened https://github.com/panther-labs/panther_analysis_tool/pull/113 to help us clear this diff.

panther_analysis_tool/rule.py
```diff
@@ -79,27 +95,35 @@ class RuleResult:
...
     runbook_output: Optional[str] = None
     runbook_exception: Optional[Exception] = None
+    runbook_defined: bool = False
 
     destinations_output: Optional[List[str]] = None
     destinations_exception: Optional[Exception] = None
+    destinations_defined: bool = False
 
-    alert_context: Optional[str] = None
+    alert_context_output: Optional[str] = None
```
georgepsarakis, line 124: Renamed for uniformity and so that we can fetch output dynamically by function name.

panther_analysis_tool/rule.py
```diff
@@ -184,25 +252,32 @@ def __init__(self, config: Mapping):
...
         """
         self.logger = get_logger()
 
         # Check for required string fields
-        for each_field in ["id", "body", "versionId"]:
+        for each_field in ["id", "versionId"]:
             if not (each_field in config) or not isinstance(config[each_field], str):
                 raise AssertionError('Field "%s" of type str is required field' % each_field)
 
+        if not (config.get("body") or config.get("path")):
```
georgepsarakis, line 267: The rule code module can now be imported directly from the filesystem or from raw source code, which is the case in Panther.

panther_analysis_tool/rule.py
```diff
@@ -289,55 +379,84 @@ def run(
             return rule_result
 
         try:
-            rule_result.title_output = self._get_title(event, use_default_on_exception=batch_mode)
+            rule_result.title_defined = self._is_function_defined("title")
```
georgepsarakis, line 382: Since the `_get_<auxiliary_function_result>` methods were returning `None`, it was not feasible to discern whether a `None` value was returned (acceptable return type for `destinations`) or the function was not defined. Now we know whether the function was defined, so we can display the output as well.

panther_analysis_tool/rule.py
```diff
@@ -289,55 +379,84 @@ def run(
...
+            rule_result.title_output = self._get_title(
+                event,
+                function_defined=rule_result.title_defined,
+                use_default_on_exception=batch_mode,
+            )
         except Exception as err:  # pylint: disable=broad-except
             rule_result.title_exception = err
 
         try:
+            rule_result.description_defined = self._is_function_defined(DESCRIPTION_FUNCTION)
```
georgepsarakis, line 392: I used constants for the function names (which can be reused in `main` as well for the list of function names), to reduce the likelihood of an error or discrepancy between the two literals.

panther_analysis_tool/rule.py
```diff
@@ -349,16 +468,16 @@ def _run_rule(self, event: PantherEvent) -> bool:
...
         if self.rule_type == TYPE_SCHEDULED_RULE and not hasattr(self._module, "rule"):
             return True
-        return self._run_command(self._module.rule, event, bool)
+        return self._run_command(self._module.rule, event, bool)  # type: ignore[attr-defined]
 
     def _get_alert_context(
-        self, event: PantherEvent, use_default_on_exception: bool = True
+        self, event: PantherEvent, function_defined: bool, use_default_on_exception: bool = True
     ) -> Optional[str]:
-        if not hasattr(self._module, "alert_context"):
```
georgepsarakis, line 357: Note that we actually need to check the attribute existence only the first time, probably once the module is loaded (?). This is a rather small optimization both in terms of flow clarity and speed, so I left it for the future.

panther_analysis_tool/rule.py
```diff
@@ -512,13 +635,16 @@ def _get_destinations(
                     str(invalid_destinations),
                 )
                 return None
-            raise ValueError("Invalid Destinations: {}".format(str(invalid_destinations)))
+            raise UnknownDestinationError(
```
georgepsarakis, line 638: Raising a specific exception type, helps to identify the error with precision and run the fallback return value check in PAT, when no available destination names are given through the CLI.

panther_analysis_tool/main.py
```diff
@@ -56,8 +59,14 @@
 )
 
 from panther_analysis_tool.data_model import DataModel
+from panther_analysis_tool.destination import FakeDestination
 from panther_analysis_tool.enriched_event import PantherEvent
+from panther_analysis_tool.exceptions import (
+    FunctionReturnTypeError,
```
georgepsarakis, line 65: For return type errors we now return a specific error type, so that we can detect and display that to the user.

panther_analysis_tool/main.py
```diff
@@ -1122,115 +1149,96 @@ def _run_tests(
...
-                is_valid = False
-                output = AttributeError(
-                    f"Expected [{function_name}] to return a list of strings, "
-                    f"got types [{list_types}] instead"
-                )
-        else:
-            is_valid = False
-    # For reserved functions besides severity and destinations, we expect a string output
+    """Determine whether an auxiliary function for a rule raised an error"""
+    function_error = getattr(rule_result, f"{function_name}_exception")
```
georgepsarakis, line 1204: I am not particularly fond of this approach, but it helps in our case.

panther_analysis_tool/main.py
```diff
@@ -1122,115 +1149,96 @@ def _run_tests(
...
-            expected_type = "list"
-        else:
-            expected_type = "string"
 
-        output = AttributeError(
-            f"Expected [{function_name}] to return a [{expected_type}], "
-            f"got [{type(function_output)}] instead"
-        )
-    return is_valid, output
+def _check_destinations_type(obj: Any) -> bool:
```
georgepsarakis, line 1233: This check should normally be provided by the `rule` module (or the `Rule` class). I didn't want to make further changes to the `Rule` class at this point to avoid making the PR more complex.

panther_analysis_tool/main.py
```diff
@@ -1294,6 +1302,15 @@ def setup_parser() -> argparse.ArgumentParser:
         "type": strtobool,
         "help": "Meant for advanced users; allows skipping of extra keys from schema validation.",
     }
+    available_destination_name = "--available-destination"
```
georgepsarakis, line 1305: Let me know if I should add documentation for the new CLI parameter!

panther_analysis_tool/main.py
georgepsarakis: Done!

georgepsarakis:
Rebased and ready for review, thank you for your patience! Also, could someone please execute a test run with this branch locally and check if the output seems different in any way? Maybe the auxiliary function name order is slightly different now, but apart from that everything should be identical.

nhakmiller:
> Rebased and ready for review, thank you for your patience! Also, could someone please execute a test run with this branch locally and check if the output seems different in any way? Maybe the auxiliary function name order is slightly different now, but apart from that everything should be identical.

I ran master and this version and diffed the results, and it appears this version is adding some new lines. On the master version:

```
<...more output...>
AWS.Redshift.Cluster.VersionUpgrade
  [PASS] Allow Version Upgrade Not Set
  [PASS] Allow Version Upgrade
  [PASS] Does Not Allow Version Upgrade

Standard.AdminRoleAssigned
  [PASS] GCP - Admin Assigned
    [PASS] [alert_context] {'ips': [], 'actor': 'bob@example.com', 'user': 'cat@example.com'}
    [PASS] [title] GCP.AuditLog: [bob@example.com] assigned admin privileges [roles/resourcemanager.organizationAdmin] to [cat@example.com]
  [PASS] GSuite - Other Admin Action
  [PASS] GSuite - Privileges Assigned
    [PASS] [alert_context] {'ips': [], 'actor': 'bobert@example.com', 'user': 'bob@example.com'}
    [PASS] [title] GSuite.Reports: [bobert@example.com] assigned admin privileges [Some Admin Role] to [bob@example.com]
  [PASS] OneLogin - Non permissions assigned event
  [PASS] OneLogin - Non super user permissions assigned
  [PASS] OneLogin - Super user permissions assigned
    [PASS] [alert_context] {'ips': [], 'actor': "Bobert O'Bobly", 'user': 'Evil Bob'}
    [PASS] [title] OneLogin.Events: [Bobert O'Bobly] assigned admin privileges [Super user] to [Evil Bob]

Standard.BruteForceByIPAWS.Redshift.Cluster.VersionUpgrade
  [PASS] Allow Version Upgrade Not Set
  [PASS] Allow Version Upgrade
  [PASS] Does Not Allow Version Upgrade

Standard.AdminRoleAssigned
  [PASS] GCP - Admin Assigned
    [PASS] [alert_context] {'ips': [], 'actor': 'bob@example.com', 'user': 'cat@example.com'}
    [PASS] [title] GCP.AuditLog: [bob@example.com] assigned admin privileges [roles/resourcemanager.organizationAdmin] to [cat@example.com]
  [PASS] GSuite - Other Admin Action
  [PASS] GSuite - Privileges Assigned
    [PASS] [alert_context] {'ips': [], 'actor': 'bobert@example.com', 'user': 'bob@example.com'}
    [PASS] [title] GSuite.Reports: [bobert@example.com] assigned admin privileges [Some Admin Role] to [bob@example.com]
  [PASS] OneLogin - Non permissions assigned event
  [PASS] OneLogin - Non super user permissions assigned
  [PASS] OneLogin - Super user permissions assigned
    [PASS] [alert_context] {'ips': [], 'actor': "Bobert O'Bobly", 'user': 'Evil Bob'}
    [PASS] [title] OneLogin.Events: [Bobert O'Bobly] assigned admin privileges [Super user] to [Evil Bob]

Standard.BruteForceByIP
<...more output...>
```

On this version:

```
<...more output...>
AWS.Redshift.Cluster.VersionUpgrade
  [PASS] Allow Version Upgrade Not Set
  [PASS] Allow Version Upgrade
  [PASS] Does Not Allow Version Upgrade

Standard.AdminRoleAssigned
  [PASS] GCP - Admin Assigned
    [PASS] [alert_context] {"ips": [], "actor": "bob@example.com", "user": "cat@example.com"}
    [PASS] [title] GCP.AuditLog: [bob@example.com] assigned admin privileges [roles/resourcemanager.organizationAdmin] to [cat@example.com]

  [PASS] GSuite - Other Admin Action
  [PASS] GSuite - Privileges Assigned
    [PASS] [alert_context] {"ips": [], "actor": "bobert@example.com", "user": "bob@example.com"}
    [PASS] [title] GSuite.Reports: [bobert@example.com] assigned admin privileges [Some Admin Role] to [bob@example.com]

  [PASS] OneLogin - Non permissions assigned event
  [PASS] OneLogin - Non super user permissions assigned
  [PASS] OneLogin - Super user permissions assigned
    [PASS] [alert_context] {"ips": [], "actor": "Bobert O'Bobly", "user": "Evil Bob"}
    [PASS] [title] OneLogin.Events: [Bobert O'Bobly] assigned admin privileges [Super user] to [Evil Bob]


Standard.BruteForceByIP
<...more output...>
```

Additionally when printing the `alert_context` it appears now we're printing with double quotes whereas before we were printing with single quotes, but that doesn't really concern. I'm more concerned about the unusual newlines, which breaks up the natural flow of reading the tests.

Specifically observe how on the old version, there is always exactly one new line between each set of test results. In the new version, sometimes there is one and sometimes there is two. Additionally, the new version sometimes inserts a new line in the middle of the test results. I suspect these two things are related.

I've attached the old & new output in case it assists with troubleshooting the difference, haven't had time to dig into why. I suspect it might be related to printing the `title` string though.
[next.txt](https://github.com/panther-labs/panther_analysis_tool/files/6811153/next.txt)
[old.txt](https://github.com/panther-labs/panther_analysis_tool/files/6811154/old.txt)

panther_analysis_tool/destination.py
nhakmiller: Getting or exporting a list would both work. My concern with having `pat` retrieve the list is we'd need to update the IAM permissions of the `pat` role for all customers. The public API can't come fast enough! For now the solution of allowing users to specify destinations with a CLI flag seems good.

nhakmiller (commented):
Big changes, but important ones!

This code looks good to me. I have a few questions, but nothing major. The only thing blocking my approval is that the output has some extraneous new lines inserted, see my previous comment. Once that has been addressed (and I can help validate!) then I am good to approve. Thank you for your hard work on this one!

panther_analysis_tool/main.py
georgepsarakis: I think repeating the CLI parameter is considered a safer approach to handle arbitrary input, because choosing a delimiter always complicates the parsing process (or the user must apply some custom character escaping). I think we currently allow the comma character in a display name, but since we already use comma-delimited input in other parameters, I'll change it to your suggestion.

panther_analysis_tool/rule.py
```diff
@@ -184,25 +252,32 @@ def __init__(self, config: Mapping):
...
         self.logger = get_logger()
 
         # Check for required string fields
-        for each_field in ["id", "body", "versionId"]:
+        for each_field in ["id", "versionId"]:
             if not (each_field in config) or not isinstance(config[each_field], str):
                 raise AssertionError('Field "%s" of type str is required field' % each_field)
 
+        if not (config.get("body") or config.get("path")):
+            raise ValueError('one of "body", "path" must be defined')
```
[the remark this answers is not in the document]
georgepsarakis: I am not sure if type checks are a usual practice in Python in general, that's why I omitted it. But I'll add it back as you mention, no harm in getting a more precise error since these are critical parameters.

panther_analysis_tool/rule.py
```diff
@@ -349,16 +468,16 @@ def _run_rule(self, event: PantherEvent) -> bool:
...
             return True
-        return self._run_command(self._module.rule, event, bool)
+        return self._run_command(self._module.rule, event, bool)  # type: ignore[attr-defined]
 
     def _get_alert_context(
-        self, event: PantherEvent, use_default_on_exception: bool = True
+        self, event: PantherEvent, function_defined: bool, use_default_on_exception: bool = True
     ) -> Optional[str]:
-        if not hasattr(self._module, "alert_context"):
+        if not function_defined:
```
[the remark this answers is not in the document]
georgepsarakis: Note that the `*_defined` attributes belong to `RuleResult` not the `Rule` class (which is the instance of the method here). I thought it would be more consistent to pass the function existence check result to the functions directly, rather than having them repeat the check.

Your comment made me realize that we could be passing the `RuleResult` instance to these functions and let them get any information they need from previous steps though. For example, `dedup` depends on `title`: https://github.com/panther-labs/panther_analysis_tool/pull/109/files#diff-acce689b4ddfb057c8806c0ce5a84fcf19b7163d8dee29fd1d4155d50716a415R447

Would that preferable / easier to follow?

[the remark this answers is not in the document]
georgepsarakis: > Another option is that since we are checking/setting the `*_defined` in the `RuleResult` outside of these functions, we could determine it isn't set/skip the `_get_*` calls altogether and not pass this in at all?

With the exception of `_get_dedup` that is possible, but I'd have to move the default dedup string / fall-back to title outside of this method. It may be a smaller change though altogether, sounds better so I'll probably do what you suggest instead.

panther_analysis_tool/main.py
```diff
@@ -1122,115 +1149,96 @@ def _run_tests(
...
+    # pass destination names explicitly through command-line parameters.
+    if function_name == "destinations" and failed and not strict_check:
+        # Otherwise we fall back to a best-effort check of the return type only
+        if isinstance(function_error, UnknownDestinationError):
+            exc: UnknownDestinationError = function_error
+            failed = not _check_destinations_type(exc.result())
+
+    # The function has not been executed
+    if not is_defined:
+        display = None
```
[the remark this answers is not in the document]
georgepsarakis: If I remember correctly, empty strings or `None` (`nil`) are omitted from the output.

panther_analysis_tool/rule.py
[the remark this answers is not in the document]
nhakmiller: Sounds good. And really I'm open to having it in both places or not having it in both places, with type hints it doesn't seem so critical I just want us to be consistent.

panther_analysis_tool/main.py
nhakmiller: I see, I didn't think about the fact that destinations may have commas in the name. Feel free to leave it as is then, otherwise we'll likely just have to revert it the first time we get a customer that puts a comma in the name.

panther_analysis_tool/rule.py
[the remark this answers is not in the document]
nhakmiller: > Note that the *_defined attributes belong to RuleResult not the Rule class

I see, that was where my confusion was coming in. Thanks for clarifying! I'll leave it to you @lindsey-w as to what makes the most sense here then.

panther_analysis_tool/main.py
georgepsarakis: Great, I'll just leave it as is then for now. I also verified by adding a destination from the UI and we do allow comma.

panther_analysis_tool/rule.py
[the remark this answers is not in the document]
georgepsarakis: Done in 53eff3c92d3943935a54a95a650d352c9b6da1c7, with tests.

georgepsarakis:
@wey-chiang @lindsey-w @nhakmiller I think I addressed all open comments in the last two commits. Summarizing the primary points:
- Check auxiliary function existence outside of `_get_<function>` executor methods. 
- Calculate dedup fallback in the main `Rule.run` method.
- Fix extra newlines in output.

Let me know what you think.

georgepsarakis:
> Additionally when printing the alert_context it appears now we're printing with double quotes whereas before we were printing with single quotes, but that doesn't really concern me.

This is because we serialize the return value of `alert_context` to JSON now:

```python
>>> a = {'ip': '8.8.8.8'}
>>> print(a)
{'ip': '8.8.8.8'}
>>> import json
>>> print(json.dumps(a))
{"ip": "8.8.8.8"}
```

Minor for sure, but I think the display is being a bit more consistent with the UI view now.

lindsey-w (approved):
🤩  a lot of stuff here - looks good to me.  Output looks good when running `make integration` 👍🏻

PR #111, opened 2021-07-03
Fix bug in test - add global_helpers to search dir

slw07g:
### Background

Fix a bug in test that causes the global_helpers directory to not get added to the search path when --path is not ".". This would cause tests to fail when rule had dependencies in global_helpers.

### Changes

Adds each `HELPERS_LOCATION` and `DATA_MODEL_LOCATION` to search_directories - if the path exists, regardless of what --path is passed to `test`. 

### Testing

- Create a rule that has a dependency in global_helpers.
- Test a specific rule (or a single directory of rules) with `panther_analysis_tool test --path path/to/rule`

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -678,8 +678,11 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
         "." + DATA_MODEL_LOCATION,
     ):
         absolute_dir_path = os.path.abspath(os.path.join(args.path, directory))
+        absolute_helper_path = os.path.abspath(directory)
         if os.path.exists(absolute_dir_path):
             search_directories.append(absolute_dir_path)
+        if os.path.exists(absolute_helper_path):
+            search_directories.append(absolute_helper_path)
 
     # First classify each file, always include globals and data models location
     specs, invalid_specs = classify_analysis(list(load_analysis_specs(search_directories)))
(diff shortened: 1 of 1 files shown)


3nvi:
Thanks @slw07g 🎉

nhakmiller:
I see, so it appears this addresses the case where:

- You need to test a rule that imports a global
- You run `panther_analysis_tool` with the `--path` flag, specifying a directory other than `.`
- You do _not_ have a `global_helpers` directory in the current working directory
- You _do_ have a `global_helpers` directory in the parent directory of the current working directory

Is that correct? I checked this branch out and that appears to be what it's doing, just want to verify I understand the exact issue this is resolving.

nhakmiller (approved):
Change looks good to me, just waiting for validation from the author about the intended use case before merging.

slw07g:
> I see, so it appears this addresses the case where:
> 
> * You need to test a rule that imports a global
> * You run `panther_analysis_tool` with the `--path` flag, specifying a directory other than `.`
> * You do _not_ have a `global_helpers` directory in the current working directory
> * You _do_ have a `global_helpers` directory in the parent directory of the current working directory
> 
> Is that correct? I checked this branch out and that appears to be what it's doing, just want to verify I understand the exact issue this is resolving.

Yes. I tried running pat from `panther_analysis` root, but one of the rules has a dependency on a custom module we have stored in `global_helpers/x_helpers/custom_mdule.py`. Adding the logic in this PR resolved the ModuleNotFound exception that was being raised.

The testing worked fine from `panther-analysis` root when `--path` is ".", but that tests all rules which is undesired. 

The cause of the issue is the current logic only looks for `global_helpers` relative to what `--path` (`args.path`) is set to.

nhakmiller:
I see now, makes sense. Good catch @slw07g!

Also it looks like some circle CI configurations weren't set properly for external contributors, I've updated them now but can you try pushing up a small change to jiggle it? Like just move line 681 down two lines or add a new line or something 😅 thanks!

PR #117, opened 2021-07-22
Fix datamodels filters

nhakmiller:
### Background

Data models don't work when a filter is provided.

### Changes

* Make data models work with filters

### Testing

* local manual testing


lindsey-w (approved):
LGTM, although it looks like an integration test is failing - I'm guessing it is missing a mock definition

nhakmiller:
> LGTM, although it looks like an integration test is failing - I'm guessing it is missing a mock definition

Good catch! @glerb and I were able to confirm that and correct it.

PR #119, opened 2021-08-02
Set session profile

nhakmiller:
### Background

It seems we're not correctly setting an environment variable when using the update-schemas command.

### Changes

* Set env variable correctly

### Testing

* None

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -397,6 +397,7 @@ def update_schemas(args: argparse.Namespace) -> Tuple[int, str]:
     # this must be called prior to setting up the client
     if args.aws_profile is not None:
         logging.info("Using AWS profile: %s", args.aws_profile)
+        session = boto3.Session(profile_name=args.aws_profile)
         set_env("AWS_PROFILE", args.aws_profile)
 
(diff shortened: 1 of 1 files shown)


panther_analysis_tool/main.py
```diff
@@ -397,6 +397,7 @@ def update_schemas(args: argparse.Namespace) -> Tuple[int, str]:
     # this must be called prior to setting up the client
     if args.aws_profile is not None:
         logging.info("Using AWS profile: %s", args.aws_profile)
+        session = boto3.Session(profile_name=args.aws_profile)
```
nhakmiller, line 400: We already do this for `upload_analysis`, but not here in `update_schemas`. Maybe we should do this as part of setting everything up so its always in place for any function that gets called?

lindsey-w: maybe a wrapper like `get_client` or something

lindsey-w (approved):

panther_analysis_tool/main.py
nhakmiller: I added a `get_client` wrapper function llike you suggested.

PR #122, opened 2021-08-07
FIX: Consolidate get_client function & fix to have user-defined schemas respect --aws-profile 

wey-chiang:
Closes https://app.asana.com/0/1200693863324520/1200778063401185/f

### Background

The `set_env` was being called in multiple different places unnecessarily and the `user_defined:Client` instantiated clients different than we do in `main.py`.

### Changes

* Moved the get_client function to the utils and removed the duplicated/redundant set_env calls.
* Changed the user_defined Client and Uploader to pass the argparse.Namespace up the stack to ensure that the lambda client used respects the aws-profile flag

### Testing

* `make test`
* `make integration`


panther_analysis_tool/main.py
```diff
@@ -463,15 +464,11 @@ def update_custom_schemas(args: argparse.Namespace) -> Tuple[int, str]:
...
-    if args.aws_profile is not None:
-        logging.info("Using AWS profile: %s", args.aws_profile)
-        set_env("AWS_PROFILE", args.aws_profile)
-
     normalized_path = user_defined.normalize_path(args.path)
     if not normalized_path:
         return 1, f"path not found: {args.path}"
 
-    uploader = user_defined.Uploader(normalized_path)
+    uploader = user_defined.Uploader(normalized_path, args)
```
lindsey-w, line 471: Since most of the other arguments are specific to main (zip/test/output directory) - wonder if we could simply pass in the aws profile here rather than the entire arg list? Or do you envision other arguments may be pertinent to the user schemas at some point in the future?

lindsey-w (commented):
LGTM! Just a minor question about the argument we are passing along.  Were you able to test before/after?

panther_analysis_tool/util.py
```diff
@@ -53,3 +56,19 @@ def store_modules(path: str, body: str) -> None:
     Path(os.path.dirname(path)).mkdir(parents=True, exist_ok=True)
     with open(path, "w") as py_file:
         py_file.write(body)
+
+
+def get_client(args: argparse.Namespace, service: str) -> boto3.client:
```
alxarch, line 61: Can we add some docstring comment explaining what this utility function does?

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -38,13 +40,14 @@ class Client:
     _LIST_SCHEMAS_ENDPOINT = "ListSchemas"
     _PUT_SCHEMA_ENDPOINT = "PutUserSchema"
 
-    def __init__(self) -> None:
+    def __init__(self, args: argparse.Namespace) -> None:
```
alxarch, line 43: nit: pass an `aws_profile=None` argument here and assign a value in `main`.

alxarch (commented):
LGTM

There seems to be an unused boto import somewhere
Some 'separation of concerns' comments about only passing `aws_profile` argument instead of importing `argparse`.

Defering approval to someone more Python-savvy

panther_analysis_tool/main.py
georgepsarakis: Alternatively, you could also (optionally so that tests are unaffected) inject the client instance directly.

panther_analysis_tool/log_schemas/user_defined.py
georgepsarakis: Either that or maybe use a slightly different name such as `options`? As you may notice the attribute name `_args` does not provide enough context that it refers to command-line arguments.

georgepsarakis (approved):
Thanks for taking the time to fix this!

lindsey-w (approved):

PR #123, opened 2021-08-16
Handle empty results response for existing schemas request

georgepsarakis:
Closes https://app.asana.com/0/1199906407795548/1200783112556400

### Background

An empty response may be returned if no custom schemas are yet stored in the backend.

### Changes

* Fall back to empty list as a default for existing schema on empty API response.

### Testing

* Tested end-to-end in conditions where an empty JSON object is returned by the API.


lindsey-w (commented):
LGTM - looks like mypy just wants an updated return type for that method

lindsey-w (approved):

slw07g:
Was coming to report this bug and it's already fixed! Thanks!

PR #124, opened 2021-08-19
Detect and ignore schema test case files

georgepsarakis:
Closes https://app.asana.com/0/1200693863324520/1200791354082308

### Background

Schema tests would be reported as malformed schema definitions, and were skipped but the error reporting results in a non zero exit code.

### Changes

* Filter discovered files by detecting test case files and omitting them from the path list of files to be examined as possible schema definitions.

### Testing

* Reproduced the error by adding a test file containing a single document, since YAML parsing would break either way for multidocument YAML files.
* Tested the changes end-to-end to verify that no error is reported.

### DIFF
--- a/panther_analysis_tool/log_schemas/user_defined.py
+++ b/panther_analysis_tool/log_schemas/user_defined.py
@@ -22,12 +22,14 @@ import json
 import logging
 import os
 from dataclasses import dataclass
+from itertools import filterfalse
 from typing import Any, Dict, List, Optional, Tuple, cast
 
 from botocore import client
 from ruamel.yaml import YAML
 from ruamel.yaml.parser import ParserError
 from ruamel.yaml.scanner import ScannerError
+from ruamel.yaml.composer import ComposerError
 
 from panther_analysis_tool.util import get_client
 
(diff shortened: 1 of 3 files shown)


alxarch (approved):
LGTM
If the purpose of this is to isolate schema files, I would propose to check for field names specific to a valid log schema file rather than excluding test files.
There could be other irrelevant YAML files (i.e. rules, manifests etc) in a directory that would also have to be excluded.

panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -348,6 +351,51 @@ def discover_files(base_path: str, patterns: Tuple[str, ...]) -> List[str]:
...
+        The list of absolute paths of files that possibly contain custom schema definitions.
+    """
+    return list(filterfalse(_contains_schema_tests, paths))
+
+
+def _contains_schema_tests(filename: str) -> bool:
+    """
+    Check if a file contains YAML document(s) that describe test cases for custom schemas.
+    :param filename: the full path for the file to be checked
+    :return:
```
lindsey-w, line 371: missing description?

lindsey-w (approved):

georgepsarakis:
> If the purpose of this is to isolate schema files, I would propose to check for field names specific to a valid log schema file rather than excluding test files.
There could be other irrelevant YAML files (i.e. rules, manifests etc) in a directory that would also have to be excluded.

I thought about this, perhaps use our published JSON schema to validate, but I didn't want to introduce the possibility of false validation errors (JSON schema validations differ per language) compared to the backend and/or `pantherlog`. Considering this command is part of a CI process, there is a chance that a silent ignore from the tool goes undetected and we fail to sync an otherwise valid schema definition. If we have more feedback about this in the future, we can certainly reconsider. For the time being, test files are naturally expected to live nearby their schema definitions, thus we need to explicitly ignore them.

PR #127, opened 2021-08-31
transition to DetectionResult

lindsey-w:
### Background

Relates to: https://app.asana.com/0/1199961856165998/1200617586832916/f

### Changes

* rename rule_* to detection_*
* return DetectionResult from rule.run()

### Testing

* `make test`
* `make integration`

### DIFF
--- a/panther_analysis_tool/detection.py
+++ b/panther_analysis_tool/detection.py
@@ -1,0 +1,190 @@
+"""
+Panther Analysis Tool is a command line interface for writing,
+testing, and packaging policies/rules.
+Copyright (C) 2020 Panther Labs Inc
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as
+published by the Free Software Foundation, either version 3 of the
+License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import logging
+import re
+import traceback
+from abc import abstractmethod
+from dataclasses import dataclass
+from pathlib import Path
+from types import ModuleType
+from typing import List, Optional
+
+from panther_analysis_tool.util import id_to_path, import_file_as_module, store_modules
+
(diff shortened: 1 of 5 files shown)


panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -398,8 +398,11 @@ def _contains_schema_tests(filename: str) -> bool:
 
     # - "input" and "logtype" are expected to be always present
     # - at least one of "result", "results" fields is required
-    return {'input', 'logtype', 'result'}.issubset(fields) or \
-           {'input', 'logtype', 'results'}.issubset(fields)
+    return {"input", "logtype", "result"}.issubset(fields) or {
```
georgepsarakis, line 401: Just a reminder that maybe we should add the formatter check in CI. I probably forgot to rerun it in this case.

panther_analysis_tool/rule.py
```diff
@@ -93,164 +93,6 @@
 )
 
 
-# pylint: disable=too-many-instance-attributes,unsubscriptable-object
```
georgepsarakis, line 96: Removing the classes from here will break development AWS deployments, since they use latest master by default (maybe we should change this). You can either freeze the commit in panther-enterprise (see `panther_config.yml`), or make the necessary change there too first and remove these in a subsequent PR.

panther_analysis_tool/detection.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+# pylint: disable=too-many-instance-attributes,unsubscriptable-object
+@dataclass
+class DetectionResult:
+    """Class containing the result of running a detection"""
+
+    detection_id: str
+    detection_severity: str
+    detection_type: Optional[str] = None
```
georgepsarakis, line 42: This field should probably be required, but I would advise to not do it now, because it will trigger another round of changes in tests 😄 .

georgepsarakis (commented):
Nice work! Is there anything else pending? I think I would approve if not for the Draft mode.

panther_analysis_tool/log_schemas/user_defined.py
lindsey-w: https://app.asana.com/0/1199906407795548/1200886091502394/f

wey-chiang (approved):
LGTM, though the issue of breaking dev deployments in `panther-enterprise` should be addressed prior to merging

lindsey-w:
> breaking dev deployments in `panther-enterprise` should be addressed prior to merging

I chatted with @georgepsarakis and I think the best I can do is merge this and the corresponding panther-enterprise PRs as closely as possible.

PR #128, opened 2021-09-01
detection run

lindsey-w:
### Background

Relates to:  https://app.asana.com/0/1199961856165998/1200617586832916/f

Part 2 to updating `pat` to handle detections more generically: update rule to inherit from a `Detection` class, where we can re-use functionality in both rule and policy engines.  

### Changes

* Move majority of rule logic into `Detection` class and rename fields to the more generic `detection_*`
* Rule inherits from `Detection`

### Testing

* `make test`
* `make integration`


panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
         self, identifier: str, body: str
     ) -> ModuleType:
         return super().from_string(identifier, body, self._tmp_dir)
+
+
+# pylint: disable=too-many-instance-attributes
+class Detection:
```
georgepsarakis, line 255: It was considerably difficult to review changes since code was moved. If it helps, I did a `git checkout` on the `rule.py` on a previous commit and diffed the two files (detection.py & rule.py) to identify the changes. Would it be too much to rename the `Rule` class to `Detection` in the `rule.py` and modify it to be more abstract? As a final step (different commit/branch?) you can move the class in `detection.py` verbatim and define the sub-class only in `rule.py`.

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+        return importer.get_module(identifier, resource)
+
+    def _check_defined_functions(self) -> Dict[str, bool]:
+        function_definitions = {}
+        for name in AUXILIARY_FUNCTIONS:
+            function_definitions[name] = self._is_function_defined(name)
+        return function_definitions
+
+    @property
+    def matcher_function(self) -> str:
```
georgepsarakis, line 359: Maybe `matcher_function_name` ? Also, you can make `Detection` an Abstract Base class (if you decide to not rename `Rule -> Detection` and proceed with inheritance in this PR) so this would become an [abstract property](https://docs.python.org/3/library/abc.html#abc.abstractproperty). Sub-classes must define these methods (the IDE would even warn you!), otherwise they cannot be instantiated, plus the abstract class cannot be instantiated too, so there's no confusion there.

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+        self, event: PantherEvent, use_default_on_exception: bool = True
+    ) -> Optional[str]:
+
+        try:
+            command = getattr(self._module, DESCRIPTION_FUNCTION)
+            description = self._run_command(command, event, str)
+        except Exception as err:  # pylint: disable=broad-except
+            if use_default_on_exception:
+                self.logger.info(
+                    "description method for rule with id [%s] raised exception. "
```
georgepsarakis, line 611:
```suggestion
                    "description method for detection with id [%s] raised exception. "
```

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+                    err,
+                )
+                return ""
+            raise
+
+        if len(description) > MAX_GENERATED_FIELD_SIZE:
+            # If generated field exceeds max size, truncate it
+            self.logger.info(
+                "maximum field [description] length is [%d]. "
+                "[%d] for rule with ID [%s] . Truncating.",
```
georgepsarakis, line 623:
```suggestion
                "[%d] for detection with ID [%s] . Truncating.",
```

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+
+        if not dedup_string:
+            # If dedup string is None or empty, return the default dedup string
+            return self._default_dedup_string
+
+        if len(dedup_string) > MAX_DEDUP_STRING_SIZE:
+            # If dedup_string exceeds max size, truncate it
+            self.logger.info(
+                "maximum dedup string size is [%d] characters. "
+                "Dedup string for rule with ID [%s] is [%d] characters. Truncating.",
```
georgepsarakis, line 583:
```suggestion
                "Dedup string for detection with ID [%s] is [%d] characters. Truncating.",
```

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+                return None
+            raise UnknownDestinationError(
+                "Invalid Destinations",
+                invalid_destinations,
+            )
+
+        if len(standardized_destinations) > MAX_DESTINATIONS_SIZE:
+            # If generated field exceeds max size, truncate it
+            self.logger.info(
+                "maximum len of destinations [%d] for rule with ID "
```
georgepsarakis, line 691:
```suggestion
                "maximum len of destinations [%d] for detection with ID "
```

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+        self, event: PantherEvent, use_default_on_exception: bool = True
+    ) -> Optional[str]:
+
+        try:
+            command = getattr(self._module, REFERENCE_FUNCTION)
+            reference = self._run_command(command, event, str)
+        except Exception as err:  # pylint: disable=broad-except
+            if use_default_on_exception:
+                self.logger.info(
+                    "reference method for rule with id [%s] raised exception. "
```
georgepsarakis, line 710:
```suggestion
                    "reference method for detection with id [%s] raised exception. "
```

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+                    err,
+                )
+                return ""
+            raise
+
+        if len(reference) > MAX_GENERATED_FIELD_SIZE:
+            # If generated field exceeds max size, truncate it
+            self.logger.info(
+                "maximum field [reference] length is [%d]. "
+                "[%d] for rule with ID [%s] . Truncating.",
```
georgepsarakis, line 722:
```suggestion
                "[%d] for detection with ID [%s] . Truncating.",
```

georgepsarakis: Same change for other log/error messages.

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+                        type(result).__name__,
+                        expected_type.__name__,
+                    )
+                )
+        else:
+            if result is None:
+                return result
+            if not isinstance(result, list) or not all(isinstance(x, (str, bool)) for x in result):
+                raise FunctionReturnTypeError(
+                    "rule [{}] function [{}] returned [{}], expected a list".format(
```
georgepsarakis, line 838:
```suggestion
                    "detection [{}] function [{}] returned [{}], expected a list".format(
```

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+            # Reports are Dict[str, List[str]]
+            # Sorting the List before setting it
+            for values in config["reports"].values():
+                values.sort()
+            self.detection_reports = config["reports"]
+
+        self._setup_exception = None
+        try:
+            self._module = self._load_detection(self.detection_id, config)
+            if not hasattr(self._module, self.matcher_function):
```
georgepsarakis, line 324: Minro, you can also use `_is_function_defined` for the check.

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+            function_definitions[name] = self._is_function_defined(name)
+        return function_definitions
+
+    @property
+    def matcher_function(self) -> str:
+        # the matcher function will differ depending on detection type
+        return "rule"
+
+    @property
+    def trigger_alert(self) -> bool:
```
georgepsarakis, line 364: Should be also abstract property?

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+        if self._setup_exception:
+            detection_result.setup_exception = self._setup_exception
+            return detection_result
+
+        try:
+            detection_result.matched = self._run_detection(event)
+        except Exception as err:  # pylint: disable=broad-except
+            detection_result.detection_exception = err
+
+        if batch_mode and detection_result.matched != self.trigger_alert:
```
georgepsarakis, line 412: `matched` may have a misleading meaning for policies? You could read it like `matched -> False` means _does not match the compliance requirements_ , though.

georgepsarakis: Truth value checking can be a bit tricky here. Although the equality comparison also works, since booleans are singletons, you can use the object identity operator `is` and maybe consider renaming the property, something like:

```python
detection_result.matched is self.matcher_alert_value
```

panther_analysis_tool/rule.py
```diff
@@ -111,559 +43,7 @@ def __init__(self, config: Mapping):
...
-                    )
-                )
-        return result
-
-    def _is_function_defined(self, name: str) -> bool:
-        return hasattr(self._module, name)
+        # initialize core parameters
+        super().__init__(config, TYPE_RULE)
+        # rule specific fields
+        self.rule_id = self.detection_id
```
georgepsarakis, line 49: This attribute isn't necessary now?

georgepsarakis (commented):
Nice work! The most important note I think here, is to check if we can make the transition changes as easy to inspect as possible.

panther_analysis_tool/detection.py
```diff
@@ -189,3 +249,597 @@ def get_module(  # pylint: disable=arguments-differ
...
+            raise ValueError('one of "body", "path" must be defined')
+
+        if config.get("body") and config.get("path"):
+            raise ValueError('only one of "body", "path" must be defined')
+
+        if not any((isinstance(config.get("body"), str), isinstance(config.get("path"), str))):
+            raise TypeError('"body", "path" parameters must be of string type')
+
+        # backwards compatible for data from before we returned this
+        self.detection_type = config.get("analysisType", default_type)
```
georgepsarakis, line 288: Ah, I think `default_type` should not be a constructor parameter. You can define the fallback here, because it might be confusing when someone sees the policy constructor and the still see `rule` in this parameter.

panther_analysis_tool/detection.py
lindsey-w: > You could read it like matched -> False means does not match the compliance requirements , though.

I'd say if a resource doesn't match the policy (`mached == False`) then it isn't compliant, so I could make an argument it makes sense in that case too 🤷🏻‍♀️

panther_analysis_tool/detection.py
lindsey-w: I guess I'm not as familiar with the backwards compatibility reqs here - were there policies that had a defined "rule" function? I removed it from constructor and created a property for the default type (so we can use `policy` as default in policies and `rule` as default in rules)

panther_analysis_tool/detection.py
lindsey-w: Yea, sorry about that - I just did what you suggested. Hopefully that makes it a little easier -_-

panther_analysis_tool/detection.py
georgepsarakis: Agreed, that's what I meant in my comment actually!

panther_analysis_tool/rule.py
```diff
@@ -659,11 +679,34 @@ def _run_command(self, function: Callable, event: PantherEvent, expected_type: A
...
 
     def _is_function_defined(self, name: str) -> bool:
         return hasattr(self._module, name)
+
+
+class Rule(Detection):
+    """Panther rule metadata and imported module."""
+
+    # default detection types for rules
+    default_detection_type = "rule"
```
georgepsarakis, line 696: Aha! I didn't know that this would satisfy the abstract property requirement, but it seems it does. The only drawback versus implementing them as property getters, is that someone can actually modify them more easily, as they seem as public class attributes. Not a really major issue though I believe, this will actually be just a bit more performant.

georgepsarakis (approved):

panther_analysis_tool/rule.py
```diff
@@ -94,20 +95,19 @@
 
 
 # pylint: disable=too-many-instance-attributes
-class Rule:
-    """Panther rule metadata and imported module."""
+class Detection:
```
georgepsarakis, line 98: Minor: if this class inherits from [ABC](https://docs.python.org/3/library/abc.html#abc.ABC), it cannot be instantiated.

nhakmiller (approved):
Abstraction ftw

PR #129, opened 2021-09-02
add policy class

lindsey-w:
### Background

Relates to:  https://app.asana.com/0/1199961856165998/1200617586832916/f

Add new policy class that can take advantage of all the function in the `Detection` class! First step towards getting dynamic fields into policies

### Changes

* Policy inherits from `Detection`

### Testing

* `make test`
* `make integration`


wey-chiang (approved):

panther_analysis_tool/policy.py
```diff
@@ -0,0 +1,47 @@
...
+    # policies have a rule method
+    matcher_function_name = "policy"
+
+    # a policy should trigger an alert on False return value
+    matcher_alert_value = False
+
+    # suppressions for the policy
+    suppressions: List[str] = []
+
+    def matcher_function(self, event: PantherEvent) -> bool:
```
nhakmiller, line 45: Hmm does it make sense to have a `PantherEvent` type describing a resource?

In the current world, I think the answer is no.

However, since the purpose of this is in support of unifying the policy & rule pipelines to a single experience, I think it makes a certain amount of sense. My only concern really is that if we add some functionality to  `PantherEvent`s that really isn't applicable to resources then that could be weird & unintuitive.

Also, I don't have an idea of what a more generic name would be. So I think `PantherEvent` is fine. Just wanted to raise the concern that we're blurring two different concepts into one here.

panther_analysis_tool/policy.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+from typing import List
+
+from panther_analysis_tool.enriched_event import PantherEvent
+from panther_analysis_tool.rule import Detection
+
+TYPE_POLICY = "POLICY"
+
+ERROR_TYPE_POLICY = "POLICY_ERROR"
```
nhakmiller, line 27: It seems that nothing is using this variable yet, should it be or is this here against future need?

nhakmiller (approved):
Wow! I'm impressed with how simple this. Way more lines of tests than source, exciting 😬

panther_analysis_tool/policy.py
lindsey-w: Yea, I was thinking converting to `PantherEvent` was useful because it makes the json immutable - but I don't know if that is necessarily needed for resources because we don't store the output in the datalake like we do matched events. I can change this to a standard `dict()` - since we also don't currently (/ won't?) use data models in policies, which is the other reason to use this type.

panther_analysis_tool/policy.py
lindsey-w: its for the futurrreeee, but a really easy thing to add back in, I'll remove it.

panther_analysis_tool/policy.py
georgepsarakis: One extra benefit is the case-insensitive lookups, which just make authoring a policy bit easier. I'm guessing though that policies are written more rarely and the resource structure is very specific, so this might not be a real gain. 

Regarding the immutability feature, it seems that the engine [depends on some of the resource fields](https://github.com/panther-labs/panther-enterprise/blob/25a70fbf323b23c7d6fcfaaf633071767d7b7246/internal/compliance/policy_engine/src/policy.py#L285) to be accurate in the result, not sure how critical that is though.

PR #131, opened 2021-09-10
add type by default to detectionresult

lindsey-w:
### Background

Add detection type by default to `DetectionResult`, similar to severity and id. 

### Changes

* add type by default
* update tests to take into account newly set field

### Testing

* `make test`

### DIFF
--- a/panther_analysis_tool/rule.py
+++ b/panther_analysis_tool/rule.py
@@ -248,7 +248,9 @@ class Detection(ABC):
         won't raise exceptions, so that an alert won't be missed.
         """
         detection_result = DetectionResult(
-            detection_id=self.detection_id, detection_severity=self.detection_severity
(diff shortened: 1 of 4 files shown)


panther_analysis_tool/rule.py
```diff
@@ -248,7 +248,9 @@ def run(
         won't raise exceptions, so that an alert won't be missed.
         """
         detection_result = DetectionResult(
-            detection_id=self.detection_id, detection_severity=self.detection_severity
+            detection_id=self.detection_id,
+            detection_severity=self.detection_severity,
+            detection_type=self.detection_type,
```
georgepsarakis, line 253: Perhaps you want to make this a mandatory field now?

wey-chiang (approved):

georgepsarakis (approved):

PR #132, opened 2021-09-15
move testing

lindsey-w:
### Background

Relates to: https://app.asana.com/0/1199961856165998/1200975330445328/f

In preparation for displaying policy dynamic field output in `pat`, we should migrate the testing logic to pat`

### Changes
* move testing.py and test_testing.py
* minor additional changes to testing.py

### Testing

* `make test`


lindsey-w:
I made a few changes to the testing.py file, here is the diff for "easier" comparison (?):

```
% diff policy_testing.py ./panther_analysis_tool/testing.py 
< from dataclasses import dataclass
10c22,23
< from typing import List, Dict, Any, Union, Optional, Tuple
---
> from dataclasses import dataclass
> from typing import Any, Dict, List, Optional, Tuple, Union
13d25
< from panther_analysis_tool.policy import TYPE_POLICY
18a31
> 
33,35c46,50
<     def new(cls,
<             output: Optional[Union[bool, str, List[str]]],
<             raw_exception: Optional[Exception] = None) -> Optional['FunctionTestResult']:
---
>     def new(
>         cls,
>         output: Optional[Union[bool, str, List[str]]],
>         raw_exception: Optional[Exception] = None,
>     ) -> Optional["FunctionTestResult"]:
52c67
<             prefix = f'{title}: '
---
>             prefix = f"{title}: "
54c69
<             prefix = ''
---
>             prefix = ""
56c71
<         return f'{prefix}{type(exc).__name__}: {exc}'
---
>         return f"{prefix}{type(exc).__name__}: {exc}"
79c94
<             return string[:length] + '...'
---
>             return string[:length] + "..."
85a101
> 
99a116
> 
118a136
> 
125a144
> 
132a152
> # pylint: disable=too-few-public-methods
146c166,169
<         matched = bool(self._detection_result.matched)
---
>         matched = (
>             bool(self._detection_result.matched)
>             == self._detection_result.detection_match_alert_value
>         )
152c175
<         if self._spec.expectations.detection:
---
>         if self._spec.expectations.detection == self._detection_result.detection_match_alert_value:
158,160c181,185
<         return self._detection_result.input_exception is None and \
<                not matched and \
<                not self._detection_result.detection_evaluation_failed
---
>         return (
>             self._detection_result.input_exception is None
>             and not matched
>             and not self._detection_result.detection_evaluation_failed
>         )
167c192
<             generic_error_title = 'Invalid event'
---
>             generic_error_title = "Invalid event"
178c203,205
<             detectionFunction=FunctionTestResult.new(self._detection_result.matched, self._detection_result.detection_exception)
---
>             detectionFunction=FunctionTestResult.new(
>                 self._detection_result.matched, self._detection_result.detection_exception
>             )
185,188c212
<         if (
<             (self._spec.expectations.detection and self._detection_result.detection_type != TYPE_POLICY) or
<             (not self._spec.expectations.detection and self._detection_result.detection_type == TYPE_POLICY)
<         ):
---
>         if self._spec.expectations.detection == self._detection_result.detection_match_alert_value:
191c215,217
<                     titleFunction=FunctionTestResult.new(self._detection_result.title_output, self._detection_result.title_exception),
---
>                     titleFunction=FunctionTestResult.new(
>                         self._detection_result.title_output, self._detection_result.title_exception
>                     ),
193c219,220
<                         self._detection_result.description_output, self._detection_result.description_exception
---
>                         self._detection_result.description_output,
>                         self._detection_result.description_exception,
196c223,224
<                         self._detection_result.reference_output, self._detection_result.reference_exception
---
>                         self._detection_result.reference_output,
>                         self._detection_result.reference_exception,
199c227,232
<                         self._detection_result.severity_output, self._detection_result.severity_exception
---
>                         self._detection_result.severity_output,
>                         self._detection_result.severity_exception,
>                     ),
>                     runbookFunction=FunctionTestResult.new(
>                         self._detection_result.runbook_output,
>                         self._detection_result.runbook_exception,
201d233
<                     runbookFunction=FunctionTestResult.new(self._detection_result.runbook_output, self._detection_result.runbook_exception),
203c235,239
<                         self._detection_result.destinations_output, self._detection_result.destinations_exception
---
>                         self._detection_result.destinations_output,
>                         self._detection_result.destinations_exception,
>                     ),
>                     dedupFunction=FunctionTestResult.new(
>                         self._detection_result.dedup_output, self._detection_result.dedup_exception
205d240
<                     dedupFunction=FunctionTestResult.new(self._detection_result.dedup_output, self._detection_result.dedup_exception),
207,208c242,244
<                         self._detection_result.alert_context_output, self._detection_result.alert_context_exception
<                     )
---
>                         self._detection_result.alert_context_output,
>                         self._detection_result.alert_context_exception,
>                     ),
218c254,256
<             genericError=FunctionTestResult.format_exception(generic_error, title=generic_error_title),
---
>             genericError=FunctionTestResult.format_exception(
>                 generic_error, title=generic_error_title
>             ),
220d257
<             # TODO: coordinate with FE for deprecation, currently defined for compatibility with the Go API
225c262
<             functions=TestResultsPerFunction(**function_results)
---
>             functions=TestResultsPerFunction(**function_results),
```

panther_analysis_tool/testing.py
```diff
@@ -0,0 +1,263 @@
...
+            detectionFunction=FunctionTestResult.new(
+                self._detection_result.matched, self._detection_result.detection_exception
+            )
+        )
+
+        # We don't include output from other functions
+        # unless the test was expected to match and trigger an alert.
+        # Even if the test fails, providing all the output provides a faster feedback loop,
+        # on possible additional failures.
+        if self._spec.expectations.detection == self._detection_result.detection_match_alert_value:
```
georgepsarakis, line 212: Maybe use the `is` operator here since we expect both values to be strictly boolean? It will be easier to reason about.

panther_analysis_tool/detection.py
```diff
@@ -37,8 +37,10 @@ class DetectionResult:
     """Class containing the result of running a detection"""
 
     detection_id: str
+    detection_match_alert_value: bool
```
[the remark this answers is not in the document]
lindsey-w: Yes, I was struggling with the best approach here becaused `matched` initially made sense to me for policies being false - but when trying to handle "how do I know if I should print anything in `pat`? ... it got messy.  

I like your suggested approach here - simplifies things greatly.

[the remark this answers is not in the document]
lindsey-w: > `not self.matcher_alert_value`

should it just default to `False` - i.e. default to non-alert?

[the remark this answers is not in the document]
georgepsarakis: > What do you think of having two things

That separates the concerns nicely and it's closer to what we have now. Not urgent, but I think we should rename `matched` to `trigger_alert` or something similar, because in the policy context (again due to inverted boolean return value) this may be confusing?

> What this doesn't solve is the issue of "should we run aux functions in this test" (i.e. should this have alerted?).

This is a testing concern, so that's a responsibility of the `testing` module to perform this check based on the `DetectionResult` attributes, since when the detection runs it should have no context about the environment it is running on.

> Would it be 'ok' to print the output of aux function when matched is true or the expected result doesn't match the actual result?

I think what you describe retains the current logic, doesn't it?

I made the following state matrixes to better understand the parameter values in each case and let me know if I'm missing something:


### Policies 

| `matched/trigger_alert` | `detection_output` | `detection_evaluation_failed` |
| --- | --- | --- |
| `True` | `False` | `False` |
| `False` | `None` | `True` |
| `False` | `True` | `False` |

#### Testing


| Should Trigger Alert | `detection_output` | Result |
| --- | --- | --- |
| `False` | `True` | _PASS_ |
| `False` | `False` | _FAIL_ |
| `False` | `None` | _FAIL_  |
| `True` | `None` | _FAIL_  |
| `True` | `False` | _PASS_ |
| `True` | `True` | _FAIL_ |

### Rules

| `matched/trigger_alert` | `detection_output` | `detection_evaluation_failed` |
| --- | --- | --- |
| `True` | `True` | `False` |
| `False` | `None` | `True` |
| `False` | `False` | `False` |

#### Testing

| Should Trigger Alert | `detection_output` | Result |
| --- | --- | --- |
| `False` | `False` | _PASS_ |
| `False` | `True` | _FAIL_ |
| `False` | `None` | _FAIL_  |
| `True` | `None` | _FAIL_  |
| `True` | `True` | _PASS_ |
| `True` | `False` | _FAIL_ |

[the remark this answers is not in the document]
lindsey-w: > This is a testing concern, so that's a responsibility of the testing module to perform this check based on the DetectionResult attributes, since when the detection runs it should have no context about the environment it is running on. 

yea, I guess I was just saying it makes the determination within testingpy harder to figure out. 

> I think what you describe retains the current logic, doesn't it?

hmm, I thought that if it expects to not trigger an alert and fails it still wouldn't run the aux function (because they wouldn't have been expected to have run). This is the current check in the the rules-engine:

```
        if self._spec.expectations.rule:
```

re the state diagram: this is super helpful. I think the only thing is rather than `Should Trigger Alert` we get `Expected Return` from the FE.  So we could potentially just make the comparison of the output with the expectation from the FE? (pseudo code):

```
 if expected != detection_output:
     return Fail
# else the expectation matched, so just return whether there was an error
 return error is None
 ```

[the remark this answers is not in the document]
Comment (georgepsarakis) [reply panther_analysis_tool/detection.py]: > hmm, I thought that if it expects to not trigger an alert and fails it still wouldn't run the aux function

The `Detection.run` function will still run the aux functions if we have a match we can't avoid this. It's just that in this case the output of these functions is not providing any meaningful feedback to the user, since they shouldn't have been executed, so we can exclude them from the test results.

> So we could potentially just make the comparison of the output with the expectation from the FE? (pseudo code):

I think you can check for error first. Also, if we retain a nullable detection output value, comparing the (always boolean) expectation with the output will always yield correct results: `False is None -> False`

File: panther_analysis_tool/testing.py [hunk 3]
```diff
@@ -0,0 +1,254 @@
...
+                self._detection_result.detection_output,
+                self._detection_result.detection_exception,
+            )
+        )
+
+        # We don't include output from other functions
+        # unless the test was expected to match and trigger an alert.
+        # If the test fails, providing all the output provides a faster feedback loop,
+        # on possible additional failures.
+        if self._detection_result.trigger_alert or (
```
georgepsarakis, line 200: Alright, now I think I get it 😄 . So my suggestion would be here to change the condition to check for the configured alert value of each detection type. Something like:

```python
  def _get_detection_alert_value(self):
    if self._detection_result.detection_type == TYPE_POLICY:
      return Policy.matcher_alert_value
   elsif self._detection_result.detection_type == TYPE_RULE:
      return Rule.matcher_alert_value
```

```python
if self._spec.expectations.detection == self._get_detection_alert_value()
```

Doesn't this cover all cases? If the test is expected to trigger an alert we should always display the aux functions results, regardless of whether it actually triggered the alert or not.

georgepsarakis (commented):
Sorry for the back and forth, it took me a while to understand the problem with the detection output expectation!

panther_analysis_tool/testing.py
```diff
@@ -0,0 +1,240 @@
...
+
+    id: Optional[str]  # pylint: disable=C0103
+    name: str
+    # The following two fields do not conform to Python's naming
+    # conventions, but the field names correspond
+    # to response attributes by API & FE.
+    # TODO: provide a field name translation step if necessary
+    detectionId: Optional[str]  # pylint: disable=C0103
+    genericError: Optional[str]  # pylint: disable=C0103
+    # TODO:
```
darwayne, line 113: do we still need this TODO? not sure what it means if we do

tests/unit/panther_analysis_tool/test_testing.py
```diff
@@ -0,0 +1,304 @@
...
+        self.assertEqual(FunctionTestResult.to_test_error(exc=exception), TestError(message='TypeError: wrong type'))
+        self.assertEqual(
+            FunctionTestResult.to_test_error(exc=exception, **dict(title='invalid type')),
+            TestError(message='invalid type: TypeError: wrong type')
+        )
+
+
+class TestTestCaseEvaluator(unittest.TestCase):
+
+
```
darwayne, line 78: `[minor]` notice you have two spaces here ... but one above, not sure which is right

darwayne (approved):
LGTM 💪🏾 , had some very minor comments

panther_analysis_tool/testing.py
lindsey-w: haha - nope! I'll remove

tests/unit/panther_analysis_tool/test_testing.py
lindsey-w: fixed

darwayne:
Sweet! 👍🏾

panther_analysis_tool/testing.py
```diff
@@ -163,14 +147,12 @@ def __init__(self, spec: TestSpecification, detection_result: DetectionResult):
 
     def _get_result_status(self) -> bool:
         """Get the test status - passing/failing"""
+        # Any error should mark the test as failing
+        if self._detection_result.errored:
```
georgepsarakis, line 151: Much better!

georgepsarakis (approved):
👏

bseb (approved):
Still wrapping my head around this codebase, but this all looks reasonable to me.

PR #133, opened 2021-09-17
update policy output

lindsey-w:
### Background

Closes: https://app.asana.com/0/1199961856165998/1201003028086559/f

With dynamic field output in policies, we should show the output for them when running `pat`.  Also, we should use the testing module to be consistent with the BE.

### Changes

* unify the way we call the run and aux methods between rules and policies
* update output: 
   * show dynamic fields for policies
   * print rule/policy output or error as well
* Update `FunctionTestResult` to have a per-function `matched` value
   * can be used to determine if we should print the detection status/output. (see "example rule / aux function throws error" below for difference in output)
   * can be used in the future for output checking on a per function basis
   

### Testing

* `make test`
* `make integration`

<details>
<summary> example policy output </summary>

```
 % panther_analysis_tool test --path ./tests/fixtures/detections/valid_analysis/policies
[INFO]: Testing analysis packs in ./tests/fixtures/detections/valid_analysis/policies

AWS.IAM.MFAEnabled
		[PASS] [policy] false
		[PASS] [dedup] defaultDedupString:AWS.IAM.MFAEnabled
		[PASS] [policy] false
		[PASS] [dedup] defaultDedupString:AWS.IAM.MFAEnabled

AWS.IAM.BetaTest
		[PASS] [policy] false
		[PASS] [dedup] defaultDedupString:AWS.IAM.BetaTest
		[PASS] [policy] false
		[PASS] [dedup] defaultDedupString:AWS.IAM.BetaTest

IAM.MFAEnabled Extra Fields
	[PASS] Root MFA not enabled triggers a violation.
		[PASS] [policy] false
		[PASS] [dedup] defaultDedupString:IAM.MFAEnabled Extra Fields

AWS.IAM.MFAEnabledGenerated
		[PASS] [policy] false
		[PASS] [title] THIS IS AN EXAMPLE TITLE
		[PASS] [dedup] THIS IS AN EXAMPLE TITLE
		[PASS] [alertContext] {"ip": "1.1.1.1"}
		[PASS] [description] THIS IS AN EXAMPLE DESCRIPTION.
		[PASS] [reference] THIS IS AN EXAMPLE REFERENCE.
		[PASS] [severity] CRITICAL
		[PASS] [runbook] THIS IS AN EXAMPLE RUNBOOK VALUE.
		[PASS] [destinations] ['ExampleDestinationName']
		[PASS] [policy] false
		[PASS] [title] THIS IS AN EXAMPLE TITLE
		[PASS] [dedup] THIS IS AN EXAMPLE TITLE
		[PASS] [alertContext] {"ip": "1.1.1.1"}
		[PASS] [description] THIS IS AN EXAMPLE DESCRIPTION.
		[PASS] [reference] THIS IS AN EXAMPLE REFERENCE.
		[PASS] [severity] CRITICAL
		[PASS] [runbook] THIS IS AN EXAMPLE RUNBOOK VALUE.
		[PASS] [destinations] ['ExampleDestinationName']

--------------------------
Panther CLI Test Summary
	Path: ./tests/fixtures/detections/valid_analysis/policies
	Passed: 4
	Failed: 0
	Invalid: 0
```

</details>

<details>

<summary> example rule / aux function throws error </summary>

Before:

```
% panther_analysis_tool test --path ./panther_analysis_tool/tmp
[INFO]: Testing analysis packs in ./panther_analysis_tool/tmp

AWS.CloudTrail.ResourceMadePublic
	[FAIL] ECR Made Public
		[FAIL] [alert_context] division by zero
		[FAIL] [description] division by zero

--------------------------
Panther CLI Test Summary
	Path: ./panther_analysis_tool/tmp
	Passed: 1
	Failed: 1
	Invalid: 0

--------------------------
Failed Tests Summary
	AWS.CloudTrail.ResourceMadePublic
		['ECR Made Public', 'ECR Made Public:alert_context', 'ECR Made Public:description']
```

After:

```
% panther_analysis_tool --debug test --path ./tmp
[INFO]: Testing analysis packs in ./tmp

AWS.CloudTrail.ResourceMadePublic
	[FAIL] ECR Made Public
		[FAIL] [rule] ZeroDivisionError: division by zero
		[PASS] [dedup] defaultDedupString:AWS.CloudTrail.ResourceMadePublic
		[FAIL] [alertContext] ZeroDivisionError: division by zero
		[FAIL] [description] ZeroDivisionError: division by zero

--------------------------
Panther CLI Test Summary
	Path: ./tmp
	Passed: 1
	Failed: 1
	Invalid: 0

--------------------------
Failed Tests Summary
	AWS.CloudTrail.ResourceMadePublic
		['ECR Made Public:rule', 'ECR Made Public:alertContext', 'ECR Made Public:description']
```

</details>

### DIFF
--- a/panther_analysis_tool/detection.py
+++ b/panther_analysis_tool/detection.py
@@ -147,9 +147,7 @@ class DetectionResult:
         """Returns whether the detection function raises an error or an import error occurred"""
         return bool(self.detection_exception or self.setup_exception)
 
-    def ignore_errors(
-        self, ignore_exception_types: Optional[List[Type[Exception]]] = None
-    ) -> None:
+    def ignore_errors(self, ignore_exception_types: Optional[List[Type[Exception]]] = None) -> None:
         """Used to ignore exceptions of particular types, used primarily in testing"""
         for exception_type in ignore_exception_types or []:
             if isinstance(self.detection_exception, exception_type):
--- a/panther_analysis_tool/detection.py
+++ b/panther_analysis_tool/detection.py
@@ -24,7 +24,7 @@ from abc import abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
 from types import ModuleType
-from typing import List, Optional
+from typing import List, Optional, Type
 
 from panther_analysis_tool.util import id_to_path, import_file_as_module, store_modules
 
@@ -146,6 +146,30 @@ class DetectionResult:
     def detection_evaluation_failed(self) -> bool:
         """Returns whether the detection function raises an error or an import error occurred"""
         return bool(self.detection_exception or self.setup_exception)
+
+    def ignore_errors(
+        self, ignore_exception_types: Optional[List[Type[Exception]]] = None
+    ) -> None:
+        """Used to ignore exceptions of particular types, used primarily in testing"""
+        for exception_type in ignore_exception_types or []:
+            if isinstance(self.detection_exception, exception_type):
+                self.detection_exception = None
+            if isinstance(self.title_exception, exception_type):
+                self.title_exception = None
+            if isinstance(self.description_exception, exception_type):
+                self.description_exception = None
+            if isinstance(self.reference_exception, exception_type):
+                self.reference_exception = None
+            if isinstance(self.severity_exception, exception_type):
+                self.severity_exception = None
+            if isinstance(self.runbook_exception, exception_type):
+                self.runbook_exception = None
+            if isinstance(self.destinations_exception, exception_type):
+                self.destinations_exception = None
+            if isinstance(self.dedup_exception, exception_type):
+                self.dedup_exception = None
+            if isinstance(self.alert_context_exception, exception_type):
+                self.alert_context_exception = None
 
 
 class BaseImporter:
--- a/panther_analysis_tool/detection.py
+++ b/panther_analysis_tool/detection.py
@@ -147,9 +147,9 @@ class DetectionResult:
         """Returns whether the detection function raises an error or an import error occurred"""
         return bool(self.detection_exception or self.setup_exception)
 
-    def ignore_errors(self, ignore_exception_types: Optional[List[Type[Exception]]] = None) -> None:
+    def ignore_errors(self, ignore_exception_types: List[Type[Exception]]) -> None:
         """Used to ignore exceptions of particular types, used primarily in testing"""
-        for exception_type in ignore_exception_types or []:
+        for exception_type in ignore_exception_types:
             if isinstance(self.detection_exception, exception_type):
                 self.detection_exception = None
             if isinstance(self.title_exception, exception_type):
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -802,7 +802,7 @@ def setup_data_models(data_models: List[Any]) -> Tuple[Dict[str, DataModel], Lis
     return log_type_to_data_model, invalid_specs
 
 
-def setup_run_tests( # pylint: disable=too-many-locals,too-many-arguments
+def setup_run_tests(  # pylint: disable=too-many-locals,too-many-arguments
     log_type_to_data_model: Dict[str, DataModel],
     analysis: List[Any],
     minimum_tests: int,
@@ -1101,7 +1101,7 @@ def run_tests(  # pylint: disable=too-many-arguments
     return failed_tests
 
 
-def _run_tests( # pylint: disable=too-many-arguments
+def _run_tests(  # pylint: disable=too-many-arguments
     analysis_data_models: Dict[str, DataModel],
     detection: Detection,
     tests: List[Dict[str, Any]],
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -36,7 +36,7 @@ from datetime import datetime
 from distutils.util import strtobool
 from fnmatch import fnmatch
 from importlib.abc import Loader
-from typing import Any, DefaultDict, Dict, Iterator, List, Set, Tuple
+from typing import Any, DefaultDict, Dict, Iterator, List, Set, Tuple, Type
 from unittest.mock import MagicMock, patch
 from uuid import uuid4
 
@@ -701,9 +701,13 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
             f"No analysis in {args.path} matched filters {args.filter} - {args.filter_inverted}"
         ]
 
+    ignore_exception_types: List[Type[Exception]] = []
+
     available_destinations: List[str] = []
     if args.available_destination:
         available_destinations.extend(args.available_destination)
+    else:
+        ignore_exception_types.append(UnknownDestinationError)
 
     destinations_by_name = {
         name: FakeDestination(destination_id=str(uuid4()), destination_display_name=name)
@@ -727,6 +731,7 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
         args.minimum_tests,
         args.skip_disabled_tests,
         destinations_by_name=destinations_by_name,
+        ignore_exception_types=ignore_exception_types,
     )
     invalid_specs.extend(invalid_detection)
 
@@ -803,6 +808,7 @@ def setup_run_tests(
     minimum_tests: int,
     skip_disabled_tests: bool,
     destinations_by_name: Dict[str, FakeDestination],
+    ignore_exception_types: List[Type[Exception]],
 ) -> Tuple[DefaultDict[str, List[Any]], List[Any]]:
     invalid_specs = []
     failed_tests: DefaultDict[str, list] = defaultdict(list)
@@ -845,6 +851,7 @@ def setup_run_tests(
             failed_tests,
             minimum_tests,
             destinations_by_name,
+            ignore_exception_types,
         )
         print("")
     return failed_tests, invalid_specs
@@ -1059,6 +1066,7 @@ def run_tests(  # pylint: disable=too-many-arguments
     failed_tests: DefaultDict[str, list],
     minimum_tests: int,
     destinations_by_name: Dict[str, FakeDestination],
+    ignore_exception_types: List[Type[Exception]],
 ) -> DefaultDict[str, list]:
 
     if len(analysis.get("Tests", [])) < minimum_tests:
@@ -1074,7 +1082,12 @@ def run_tests(  # pylint: disable=too-many-arguments
         return failed_tests
 
     failed_tests = _run_tests(
-        analysis_data_models, detection, analysis["Tests"], failed_tests, destinations_by_name
+        analysis_data_models,
+        detection,
+        analysis["Tests"],
+        failed_tests,
+        destinations_by_name,
+        ignore_exception_types,
     )
 
     if minimum_tests > 1 and not (
@@ -1094,6 +1107,7 @@ def _run_tests(
     tests: List[Dict[str, Any]],
     failed_tests: DefaultDict[str, list],
     destinations_by_name: Dict[str, FakeDestination],
+    ignore_exception_types: List[Type[Exception]],
 ) -> DefaultDict[str, list]:
 
     for unit_test in tests:
@@ -1134,9 +1148,7 @@ def _run_tests(
             mocks=unit_test.get("Mocks", {}),
             expectations=TestExpectations(detection=unit_test["ExpectedResult"]),
         )
-        ignore_exception_types: List[Exception] = []
-        if not bool(destinations_by_name):
-            ignore_exception_types.append(UnknownDestinationError)
+
         test_result = TestCaseEvaluator(spec, result).interpret(
             ignore_exception_types=ignore_exception_types
         )
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -1165,7 +1165,9 @@ def _check_destinations(test_result: TestResult, destination_by_name: Dict[str, 
     return test_result
 
 
-def _print_test_result(detection: Detection, test_result: TestResult, failed_tests: DefaultDict[str, list]) -> None:
+def _print_test_result(
+    detection: Detection, test_result: TestResult, failed_tests: DefaultDict[str, list]
+) -> None:
     status_pass = "PASS"  # nosec
     status_fail = "FAIL"
     if test_result.passed:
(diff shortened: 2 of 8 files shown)


panther_analysis_tool/main.py
```diff
@@ -1106,124 +1106,94 @@ def _run_tests(
...
-        return False
+    if (
+        test_result.functions.destinationsFunction
+        and test_result.functions.destinationsFunction.error
+        and not bool(destination_by_name)
+    ):
 
-    return all(isinstance(destination_name, str) for destination_name in obj)
+        error_message = test_result.functions.destinationsFunction.error.message or ""
+        if "UnknownDestinationError" in error_message:
```
lindsey-w, line 1155: This seems very hack-y, so definitely open to better way to support this backward compatibility.

tests/fixtures/detections/example_set_duplicates.yml
```diff
@@ -17,8 +17,8 @@ SummaryAttributes:
   - example_dupe
   - example_dupe
 OutputIds:
-  - 1234
-  - 1234
+  - "1234"
+  - "1234"
```
lindsey-w, line 21: This was actually causing this to be invalid because of a type check rather than duplicate set check.

tests/fixtures/detections/valid_analysis/policies/example_policy_beta.yml
```diff
@@ -9,7 +9,6 @@ ResourceTypes:
   - AWS.IAM.RootUser
   - AWS.IAM.User
 Tags:
-  - AWS Managed Rules - Security, Identity & Compliance
```
lindsey-w, line 12: Just removed to cleanup a bit (this is a policy)

panther_analysis_tool/testing.py
```diff
@@ -41,12 +41,15 @@ class FunctionTestResult:
     # error contains a TestError instance with the error message or
     # None if no error was raised
     error: Optional[TestError]
+    # return value matched expectations
+    matched: Optional[bool]
```
lindsey-w, line 45: looking for input on this approach for checking per-function matched value (helps in deciding to print the policy/rule output since the overall `passed` value can be "false" even if the rule itself has passed, for aux function errors, etc.)

lindsey-w:
The single integration test failing is because the `ImmutableList` type is failing the `isinstance(maybe_list, list)` check ([here](https://github.com/panther-labs/panther-analysis/blob/master/aws_rds_policies/aws_rds_instance_snapshot_public_access.py#L6)).  The `ImmutableList` is a `Sequence` type, so I am going to transition this check to `isinstance(maybe_list, Sequence)` - since the " x in y" check should work for all `Sequence` types (including lists, tuples, immutablelists...)

panther_analysis_tool/main.py
georgepsarakis: I don't think you can do this check from the `TestResult` attributes, you need access to `DetectionResult` which holds the exception instance.

georgepsarakis: How about `TestCaseEvaluator.interpret` handled this internally? Perhaps adding an argument:

```python
def interpret(self, ignore_exception_types=None):
```

You can then check in `interpret` that an exception in a `DetectionResult` attribute (since you have access in there) is an instance of the given types and adjust the test result / function output accordingly.

How does that sound?

panther_analysis_tool/main.py
```diff
@@ -1106,124 +1106,104 @@ def _run_tests(
...
+    status_fail = "FAIL"
+    if test_result.passed:
+        outcome = status_pass
+    else:
+        outcome = status_fail
+    # print overall status for this test
+    print("\t[{}] {}".format(outcome, test_result.name))
+
+    # print function output and status as necessary
+    for function_name in vars(test_result.functions):
```
georgepsarakis, line 1182: Since `functions` is a dataclass you can also use [asdict](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) I think. `vars` is a bit obscure I believe.

panther_analysis_tool/testing.py
georgepsarakis: Isn't this addressed in https://github.com/panther-labs/panther_analysis_tool/pull/132 ?

panther_analysis_tool/main.py
```diff
@@ -1106,124 +1106,104 @@ def _run_tests(
                     for each_mock in mocks
                     if "objectName" in each_mock and "returnValue" in each_mock
                 }
-            if is_policy:
-                # Policies use plain dict objects as input
-                test_case = entry
-            else:
-                # Set up each test case, including any relevant data models
-                test_case = PantherEvent(entry, analysis_data_models.get(log_type))
+            test_case = PantherEvent(entry, analysis_data_models.get(log_type))
```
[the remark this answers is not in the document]
lindsey-w: Yes - decided to go with `PantherEvent` rather than `dict` object

panther_analysis_tool/testing.py
lindsey-w: Well, this specifically allows us to also compare the output of the function with the expected ([here](https://github.com/panther-labs/panther_analysis_tool/pull/133/files#diff-4c3406e46ca13890fa64d295f86ed01e0b0bcd3d28031f1fc97978bb8fca2da3R1199)).  The PR you link allows us to determine whether or not the overall status of the test pass/failed.  But this allows us to print the output when it doesn't match the expected.  Today, this is only useful for the detections themselves, but I could see it being useful for the other functions in the future too (e.g. allow people to specify an expected output from the title function? so we could say the title function failed if there is an error _or_ the expected output doesn't match) Specifically allows us to print this: 

```
AWS.CloudTrail.ResourceMadePublic
	[FAIL] ECR Made Public
		[FAIL] [rule] ZeroDivisionError: division by zero
		[PASS] [dedup] defaultDedupString:AWS.CloudTrail.ResourceMadePublic
		[FAIL] [alertContext] ZeroDivisionError: division by zero
		[FAIL] [description] ZeroDivisionError: division by zero
```

rather than this:

```
AWS.CloudTrail.ResourceMadePublic
	[FAIL] ECR Made Public
		[FAIL] [alert_context] division by zero
		[FAIL] [description] division by zero
```

panther_analysis_tool/main.py
lindsey-w: Made a pass at this implementation. Didn't know if I should support passing the ignored exceptions all the way down the stack (perhaps sometime in the future we may support passing this in as a command line option?) - but I decided to do that anyway.

darwayne (approved):
you have some merge conflicts, but other than that LGTM.

panther_analysis_tool/testing.py
lindsey-w: I guess what was addressed in the other PR is being able to tell if the overall test pass/failed.  The test result has these fields:

```python
    detectionId: Optional[str]
    genericError: Optional[str]
    error: Optional[TestError]
    errored: bool
    passed: bool
    trigger_alert: Optional[bool]
    functions: TestResultsPerFunction
```

Which doesn't necessarily tell us if the detection output == expectations.  Since we don't know that from these fields, I'm unsure how we would determine if we should print the error/output from a detection?

georgepsarakis: > but I could see it being useful for the other functions in the future too (e.g. allow people to specify an expected output from the title function? 

Ah nice, it would be useful in this case!

panther_analysis_tool/testing.py
```diff
@@ -164,19 +164,50 @@ def _get_generic_error_details(self) -> Tuple[Optional[Exception], Optional[str]
...
+    def interpret(
+        self, ignore_exception_types: Optional[List[Type[Exception]]] = None
+    ) -> TestResult:
         """Evaluate the detection result taking into account
         the errors raised during evaluation and
         the test specification expectations"""
+
+        # first, we should update the detection result, taking into account any
+        # ignored exception types passed into this test
+        self._check_exception_types(ignore_exception_types)
```
georgepsarakis, line 204: We need to mutate the detection result because `DetectionResult.errored` would falsely return `True` otherwise, right?

How about if we moved the function that nullifies errors to `DetectionResult`, e.g. `DetectionResult.ignore_errors(exception_types)`. Such a function is expected to mutate the detection result instance, and it is quite easier to test as well!

lindsey-w: > We need to mutate the detection result because DetectionResult.errored would falsely return True otherwise, right?

yea, exactly 😕

panther_analysis_tool/main.py
```diff
@@ -1106,124 +1122,84 @@ def _run_tests(
...
+        if function_result:
+            if function_result.get("error"):
+                # add this as output to the failed test spec as well
+                failed_tests[detection.detection_id].append(f"{test_result.name}:{printable_name}")
+                print(
+                    "\t\t[{}] [{}] {}".format(
+                        status_fail, printable_name, function_result.get("error", {}).get("message")
+                    )
+                )
+            # if it didn't error, we simiply need to check if the output was as expected
```
georgepsarakis, line 1189:
```suggestion
            # if it didn't error, we simply need to check if the output was as expected
```

tests/unit/panther_analysis_tool/test_rule.py
```diff
@@ -755,6 +755,45 @@ def test_detection_evaluation_failed(self) -> None:
         self.assertTrue(DetectionResult(detection_id='failed.rule', detection_severity='INFO', detection_type=TYPE_RULE, setup_exception=TypeError(), trigger_alert=False).detection_evaluation_failed)
         self.assertFalse(DetectionResult(detection_id='failed.rule', detection_severity='INFO', detection_type=TYPE_RULE, title_exception=TypeError(), trigger_alert=False).detection_evaluation_failed)
 
+    def test_ignore_errores(self) -> None:
```
georgepsarakis, line 758:
```suggestion
    def test_ignore_errors(self) -> None:
```

tests/unit/panther_analysis_tool/test_rule.py
```diff
@@ -755,6 +755,45 @@ def test_detection_evaluation_failed(self) -> None:
...
+            detection_severity='INFO',
+            detection_type=TYPE_RULE,
+            detection_exception=TypeError(),
+            trigger_alert=False,
+        )
+        result.ignore_errors([TypeError])
+        self.assertFalse(result.errored)
+        self.assertIsNone(result.detection_exception)
+
+        # type error should be ignored, but the ZeroDivisonError should not
```
georgepsarakis, line 771:
```suggestion
        # type error should be ignored, but the ZeroDivisionError should not
```

panther_analysis_tool/detection.py
```diff
@@ -147,6 +147,28 @@ def detection_evaluation_failed(self) -> bool:
         """Returns whether the detection function raises an error or an import error occurred"""
         return bool(self.detection_exception or self.setup_exception)
 
+    def ignore_errors(self, ignore_exception_types: Optional[List[Type[Exception]]] = None) -> None:
+        """Used to ignore exceptions of particular types, used primarily in testing"""
+        for exception_type in ignore_exception_types or []:
```
georgepsarakis, line 152: Minor: `isinstance` can accept an iterable as a second argument, the loop is not required.

panther_analysis_tool/detection.py
```diff
@@ -147,6 +147,28 @@ def detection_evaluation_failed(self) -> bool:
         """Returns whether the detection function raises an error or an import error occurred"""
         return bool(self.detection_exception or self.setup_exception)
 
+    def ignore_errors(self, ignore_exception_types: Optional[List[Type[Exception]]] = None) -> None:
```
georgepsarakis, line 150: Sorry I missed that in the previous review, it would be better if the parameter is not optional now, especially since this function is never called without parameters or with `None`.

panther_analysis_tool/main.py
```diff
@@ -1106,124 +1122,84 @@ def _run_tests(
...
-            exc: UnknownDestinationError = function_error
-            failed = not _check_destinations_type(exc.result())
-
-    # The function has not been executed
-    if not is_defined:
-        display = None
+def _print_test_result(
+    detection: Detection, test_result: TestResult, failed_tests: DefaultDict[str, list]
+) -> None:
+    status_pass = "PASS"  # nosec
```
georgepsarakis, line 1164: Out of curiosity, why the `# nosec` directive?

georgepsarakis (approved):

panther_analysis_tool/detection.py
lindsey-w: I think it has to be a tuple rather than a list (?) - so I can convert the list to a tuple in the check or make the change to make the list a tuple everywhere

panther_analysis_tool/main.py
lindsey-w: This is for a [bandit](https://pypi.org/project/bandit/) check:

```
Test results:
>> Issue: [B105:hardcoded_password_string] Possible hardcoded password: 'PASS'
   Severity: Low   Confidence: Medium
   Location: panther_analysis_tool/main.py:1164
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b105_hardcoded_password_string.html
1163    ) -> None:
1164        status_pass = "PASS"
1165        status_fail = "FAIL"
```

panther_analysis_tool/testing.py
bseb: This is going to be extraordinarily helpful in debugging rules, very nice

wey-chiang: The ability to test the output of generated alert fields is an fx request from some customers

https://app.asana.com/0/1200908948600035/1200945739799345/f

bseb (approved):

tests/fixtures/detections/valid_analysis/policies/example_policy_generated_functions.py
```diff
@@ -0,0 +1,33 @@
...
+    return ["ExampleDestinationName"]
+
+def runbook(resource):
+    return 'THIS IS AN EXAMPLE RUNBOOK VALUE.'
+
+def reference(resource):
+    return 'THIS IS AN EXAMPLE REFERENCE.'
+
+def severity(resource):
+    return 'CRITICAL'
```
wey-chiang, line 33: `severity` should be case insensitive and that logic should be handled behind the scenes. It'd be good to make sure that's working as intended.

```suggestion
    return 'CrItIcAl'
```

wey-chiang (approved):
LGTM!

PR #139, opened 2021-09-27
Add Option to ignore files 

bseb:
### Background

Add Option to ignore files not related to Panther ( CI config and other yaml files) so PAT doesnt try to run their non-existent tests

### Changes
Added an `ignore-files` option to the `load_analysis_specs` function

Took this approach rather than extending `filter` since filter primary acts upon the yaml schema, and since these files wont have that schema it made more sense to not load them at all.

### Testing

* Unit Tests and CI

[asciicast](https://asciinema.org/a/W1Ng940SGIhyWPW3oA3abAcTm)


README.md
```diff
@@ -10,7 +10,6 @@
 </p>
 
 <p align="center">
-  <a href="https://slack.runpanther.io/">Chat with us on Slack!</a>
```
lindsey-w, line 13: 🙏🏻

panther_analysis_tool/main.py
```diff
@@ -179,17 +180,22 @@ def load_module(filename: str) -> Tuple[Any, Any]:
     return module, None
 
 
-def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any, Any]]:
+def load_analysis_specs(directories: List[str],
+                        ignored_files: OPT[list] = None) ->\
+        Iterator[Tuple[str, str, Any, Any]]:
```
lindsey-w, line 185: did `make fmt` do this? I thought `black` didn't like multilines with `\` 
```suggestion
                        ignored_files: OPT[list] = None
                        ) -> Iterator[Tuple[str, str, Any, Any]]:
```

panther_analysis_tool/main.py
```diff
@@ -1306,6 +1323,7 @@ def setup_parser() -> argparse.ArgumentParser:
     test_parser.add_argument(min_test_name, **min_test_arg)
     test_parser.add_argument(path_name, **path_arg)
     test_parser.add_argument(ignore_extra_keys_name, **ignore_extra_keys_arg)
+    test_parser.add_argument(ignore_files_name, **ignore_files_arg)
```
lindsey-w, line 1326: wonder if we need to add this to the `publish` and `upload` parsers as well.  Can you try running through all the commands?

panther_analysis_tool/main.py
```diff
@@ -179,17 +180,22 @@ def load_module(filename: str) -> Tuple[Any, Any]:
...
         directories: The relative path to Panther policies or rules.
+        ignored_files: Files that Panther Analysis Tool should not process
 
     Yields:
         A tuple of the relative filepath, directory name, and loaded analysis specification dict.
     """
     # setup a list of paths to ensure we do not import the same files
     # multiple times, which can happen when testing from root directory without filters
+    if ignored_files is None:
+        ignored_files = []
```
lindsey-w, line 198: if you add a `default=""` to the parser... 1. would it work? 2. could we get rid of this?

panther_analysis_tool/main.py
bseb: Nah this was PyCharm being "helpful"  I need to play with the settings at some point to see if it can use black's standards instead of stuff like this. 

Fixed with a `make fmt`

panther_analysis_tool/main.py
bseb: It seems no. I tried the default as an empty string and an empty list and both still ended up being `None` without this snippet. It seems the only way to drop this is to make the default argument an empty list which goes against the "no mutable defaults" standard.

panther_analysis_tool/main.py
bseb: Good catch, `upload` did for sure. `publish` doesn't require it as it doesn't call the `load_analysis_specs` function at all in the workflow, at least not that I could find. I did test all the commands after adding to `upload` and everything seems to work well.

panther_analysis_tool/main.py
georgepsarakis: You can add an [empty list as default in the parser argument definition](https://github.com/panther-labs/panther_analysis_tool/pull/139/files#diff-4c3406e46ca13890fa64d295f86ed01e0b0bcd3d28031f1fc97978bb8fca2da3R1284) and make `ignored_files` a required argument for this function.

bseb: 🤦🏼  I am not sure why that didn't occur to me. Changed accordingly.

panther_analysis_tool/main.py
```diff
@@ -223,8 +227,8 @@ def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any,
                     logging.debug("Skipping path %s", relative_path)
                     continue
             for filename in sorted(file_list):
-                # Skip hidden files
-                if filename.startswith("."):
+                # Skip hidden files and files explicitly ignored
+                if filename.startswith(".") or filename in ignored_files:
```
georgepsarakis, line 231: Hmm are you sure about this? Wouldn't this result in ignoring every file `test.yml` regardless of the directory? I think it is safer to require the relative path from the location where the executable runs. So if a user ignores `test.yml`, they actually mean `ignore "$PWD/test.yml"`.

Also I'm not sure where we stand on logging and exceptions, but consider:
- printing a warning log when encountering an ignored path
- raising an error if you don't encounter it, since that's an indication of a problem as well (mistyped path perhaps)

bseb: Agreed on both counts. I thought an INFO log might be more appropriate though. 

I intentionally did not raise an error if its not found because the next step on this will be a config file to read in args and I didn't want peoples CI jobs to fail because they removed a file and forgot to update the config file. All this config file is coming on another PR

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -56,20 +56,20 @@ def tearDown(self) -> None:
                 os.remove(os.path.join(_DATAMODEL_FOLDER, os.path.split(data_model_module)[-1]))
 
     def test_valid_json_policy_spec(self):
```
lindsey-w, line 58: Could you add a similar test that validates when passing in one+ ignore_files, they don't show up in the loaded specs?

panther_analysis_tool/main.py
lindsey-w: I think it might, call trace I see:

publish_release-> setup_release -> generate_release_assets -> zip_analysis -> load_analysis_specs

lindsey-w (commented):
Looking good to me

panther_analysis_tool/main.py
lindsey-w: It would be neat if we could support fnmatch

panther_analysis_tool/main.py
```diff
@@ -1263,6 +1276,16 @@ def setup_parser() -> argparse.ArgumentParser:
         "type": strtobool,
         "help": "Meant for advanced users; allows skipping of extra keys from schema validation.",
     }
+    ignore_files_name = "--ignore-files"
+    ignore_files_arg: Dict[str, Any] = {
+        "required": False,
+        "dest": "ignored_files",
+        "nargs": "+",
+        "help": "Relative path to files in this project to be ignored by panther-analysis tool, "
+        + "space separated. Example ./foo.yaml ./bar/baz.yaml",
```
georgepsarakis, line 1285: Not a suggestion for change, the `+` is not actually required for string literals.

panther_analysis_tool/main.py
```diff
@@ -179,11 +179,14 @@ def load_module(filename: str) -> Tuple[Any, Any]:
     return module, None
 
 
-def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any, Any]]:
+def load_analysis_specs(
+    directories: List[str], ignored_files: List[list]
```
georgepsarakis, line 183: Isn't `ignored_files` a list of strings?

```suggestion
    directories: List[str], ignored_files: List[str]
```

panther_analysis_tool/main.py
```diff
@@ -230,6 +233,11 @@ def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any,
                 # skip loading files that have already been imported
                 if spec_filename in loaded_specs:
                     continue
+                # Dont load files that are explictly ignored
```
georgepsarakis, line 236: As a note for a potential future enhancement, you could consider splitting the file discovery logic from the YAML parsing logic.

Also, I just noticed that we handle JSON files here. Normally YAML parsers can handle the JSON format, because YAML is a superset of JSON (again just a note for future consideration, completely unrelated to this PR 😄 ).

panther_analysis_tool/main.py
```diff
@@ -230,6 +233,11 @@ def load_analysis_specs(directories: List[str]) -> Iterator[Tuple[str, str, Any,
...
+                relative_name = os.path.join(relative_path, filename)
```
georgepsarakis, line 237: So if I remember correctly about what `os.walk` returns this will yield a path like `./test.yml`. The comparison will fail if the user has entered `test.yml`, right? Maybe you want to pass both paths through [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath) and then compare?

georgepsarakis (commented):
Apart from one last note regarding path normalization that you may want to consider, it looks good to me. I'll leave final approval to the other reviewers.

lindsey-w (approved):
LGTM!

PR #140, opened 2021-09-30
revert to dict

lindsey-w:
### Background

Revert policies back to dictionary type rather than `PantherEvent` until we have a migration path / pack support in the backend to update the necessary detections.

### Changes

* policies are now dictionaries rather than PantherEvent type

### Testing

* `make test`
* `make integration`


panther_analysis_tool/main.py
```diff
@@ -61,7 +62,7 @@
 from panther_analysis_tool.enriched_event import PantherEvent
```
bseb, line 62: Should we drop this import as well?

bseb (approved):
One question, but not a blocker from merging.

Pretty Straightforward, LGTM!

panther_analysis_tool/main.py
bseb: Oh never mind, I missed the other usage

PR #141, opened 2021-10-07
Add Config file parsing for options

bseb:
### Background
Creates framework for parsing config file in Yaml. This is initially scoped to the `ignore-files` option but can be expanded to other command line arguments with some testing and possibly some minor logic changes to the options. I intend to create Asana tasks for expanding it to other options, but wanted to start here in order to not create a giant PR

### Changes
- Add Dynaconf module
-  Create functions to instantiate Dynaconf object and merge Dynaconf and argparse config

### Testing

* Tested locally manually:
   * Detects presence of config file and reads it in if present
   *  Accepts list of files from config file and ignores them
   * Merges files that are ignored via the command line and files in config file
   * Honors other options called from command line that are not in config file
* Unit / CI tests

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -42,7 +42,6 @@ from uuid import uuid4
 
 import botocore
 import requests
-import semver
 from dynaconf import Dynaconf, Validator
 from ruamel.yaml import YAML
 from ruamel.yaml import parser as YAMLParser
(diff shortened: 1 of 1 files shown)


bseb:
After talking with @lindsey-w  yesterday we decided that making the config file behavior an override instead of a merge made more sense and GREATLY simplifies adding this feature. I have updated the example yaml file accordingly and merge function accordingly. I have been able to test a variety of combinations and everything seems to work as expected.

The syntax for `filter` is a little wonky, but the example yaml file has an example and brief description that should clarify.

cc: @georgepsarakis

PR #148, opened 2021-10-24
Unit Test Mocking - Native JSON Support

wey-chiang:
---

<div align="center"> <h1>WARNING<br/><strike>THIS WILL BREAK EXISTING CI/CD PIPELINES THAT USE PAT</strike>
</h1> </div>

~The `--use-legacy-mocking` flag has been added as a workaround which users will need to add to their CI/CD pipelines. 
Once users have updated their detections to align with the new unit test mocking, this flag should then be removed.~

The `--use-legacy-mocking` flag has been inverted, requiring users to explicitly opt-in to the new functionality.

Because of this, these changes constitute **at minimum** a minor version bump. 
We may want to consider these changes as warranting a major version bump.

---

### Background

Unit Test Mocking as it exists today has a limitation of only returning a string. 
Consequently, users are forced to write detections in a way that specifically handles the returned string. 
This PR implements support for native JSON object conversion and returning the Python equivalent.
These changes are required before the changes on the Panther BE can fully support native JSON.

### Changes
 * Added support for legacy mocking in PAT
 * Added flag to allow for legacy mocking
 * Added b/w compatibility for legacy mocking in the Detection object
 * Added unit test detection for legacy mocks

### Testing

* `make test`


panther_analysis_tool/main.py
```diff
@@ -1179,6 +1183,32 @@ def _run_tests(  # pylint: disable=too-many-arguments
...
+        logging.info("Using legacy mocking for %s:%s}", detection_id, test_name)
+    else:
+        for each_mock in mocks:
+            if "objectName" in each_mock and "returnValue" in each_mock:
+                try:
+                    return_object = json.loads(each_mock["returnValue"])
+                    mock_methods[
+                        each_mock["objectName"]] = MagicMock(return_value=return_object)
+                except Exception:  # pylint: disable=broad-except
+                    logging.warning("Invalid JSON Mock for %s:%s", detection_id, test_name)
```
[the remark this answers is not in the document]
wey-chiang: Changed to log the exception and raise

wey-chiang:
> Invert the feature flag semantics, by adding a flag --use-json-mocking or similar, so someone can opt-in to the new functionality.

I'll just change the default value of the `--use-legacy-mocking` to be `True` to avoid renaming the flag in the future while also requiring users to explicitly opt-in to the new functionality.

panther_analysis_tool/main.py
```diff
@@ -1179,6 +1183,32 @@ def _run_tests(  # pylint: disable=too-many-arguments
...
+        }
+        logging.info("Using legacy mocking for %s:%s}", detection_id, test_name)
+    else:
+        for each_mock in mocks:
+            if "objectName" in each_mock and "returnValue" in each_mock:
+                try:
+                    return_object = json.loads(each_mock["returnValue"])
+                    mock_methods[
+                        each_mock["objectName"]] = MagicMock(return_value=return_object)
+                except Exception:  # pylint: disable=broad-except
```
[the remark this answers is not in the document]
wey-chiang: This was intentionally broad to provide users a more informative logging message telling them which detection and unit test caused the exception

panther_analysis_tool/main.py
```diff
@@ -1179,6 +1183,32 @@ def _run_tests(  # pylint: disable=too-many-arguments
...
+        mock_methods = {
+            each_mock["objectName"]: MagicMock(return_value=each_mock["returnValue"])
+            for each_mock in mocks if "objectName" in each_mock and "returnValue" in each_mock
+        }
+        logging.info("Using legacy mocking for %s:%s}", detection_id, test_name)
+    else:
+        for each_mock in mocks:
+            if "objectName" in each_mock and "returnValue" in each_mock:
+                try:
+                    return_object = json.loads(each_mock["returnValue"])
```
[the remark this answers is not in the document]
wey-chiang: Nice catch, that's way cleaner

wey-chiang:
This PR should be good to go now

panther_analysis_tool/rule.py
```diff
@@ -133,6 +139,18 @@ def __init__(self, config: Mapping):
         self.detection_id = config["id"]
         self.detection_version = config["versionId"]
 
+        # backwards compatible for legacy mocking
+        self._setup_exception = None
+        if "lastUpdatedVersion" in config:
+            try:
+                detection_panther_version = semver.VersionInfo.parse(config["lastUpdatedVersion"])
```
lindsey-w, line 146: is there a corresponding BE PR that populates this version field?

panther_analysis_tool/rule.py
```diff
@@ -133,6 +139,18 @@ def __init__(self, config: Mapping):
...
+            except Exception as err:  # pylint: disable=broad-except
+                self._setup_exception = err
+                return
+            self.use_legacy_mocking = detection_panther_version < LEGACY_MOCKING_VERSION
+        else:
+            self.use_legacy_mocking = True
```
lindsey-w, line 152: Is there a reason to set this mocking field in the detection rather than outside the detection? (i.e. set the `detection_panther_version` here and just use that in `_resolve_mocks ` method?)

panther_analysis_tool/schemas.py
```diff
@@ -134,6 +134,7 @@
                 Optional("Mocks"): object,
             }
         ],
+        Optional("LastUpdatedVersion"): str,
```
lindsey-w, line 137: Do we need this here? Is the workflow for a user to pass in the cli option (vs updating their metadata files)? or are we supporting both?

further, would we want to allow specifying this in yml files? or is this something we should control based on the panther version on the backend? what if a user specifies a version and when uploading it doesn't match the panther version? do we take the one specified or force it to the panther version?

panther_analysis_tool/rule.py
wey-chiang: Not currently, will be rewriting the BE PR as a lot of the code base has changed since I initially drafted that PR

panther_analysis_tool/schemas.py
wey-chiang: > Do we need this here?

Yes, otherwise our unit tests will fail due to schema validation.

> Is the workflow for a user to pass in the cli option (vs updating their metadata files)? or are we supporting both?

We are supporting both -- from v1.0.0 and after, users will have to have their detection metadata files updated for the new mocking or have the `--use-legacy-mocks` flag specified. However that being said, once users have their detections updated to support the new mocking, they will have to have the `LastUpdatedVersion` added to the detection metadata. We should include these instructions in the migration notes.


> or is this something we should control based on the panther version on the backend?  what if a user specifies a version and when uploading it doesn't match the panther version? do we take the one specified or force it to the panther version?

This will be controlled on the BE. In the case where a user specifies a version that doesn't match the panther version, we'll just overwrite that version unless the version is less than the minimum version required to support the new unit test mocking.

panther_analysis_tool/rule.py
wey-chiang: No particular reason, I just figured it'd be more readable if we were to implement the semver comparison here

panther_analysis_tool/schemas.py
wey-chiang: So for existing detections, if `LastUpdatedVersion` is `nil` then we won't update the version and it'll be processed as a legacy mock. This comes with the caveat that users attempting to update detections will be required to update their unit test mocks to adhere to the new unit test mocking. 

To alleviate the burden of having customers update all of their detections after upgrading to v1.26.0, we could include a dummy version when the `--use-legacy-mocks` flag is `True` as to avoid breaking existing CI/CD pipelines. I'll push up changes to include that feature tomorrow

panther_analysis_tool/main.py
```diff
@@ -1303,6 +1333,14 @@ def setup_parser() -> argparse.ArgumentParser:
         "help": "A destination name that may be returned by the destinations function. "
         "Repeat the argument to define more than one name.",
     }
+    # TODO(wey-chiang): Change default to False in v1.0.0
+    use_legacy_mocking_name = "--use-legacy-mocking"
+    use_legacy_mocking_arg: Dict[str, Any] = {
+        "action": "store_true",
+        "default": True,
```
georgepsarakis, line 1340: I'm not sure I follow the logic here. If someone provides the flag, the value would be `True`, but the default is also `True`, so it would have no effect. I assume that currently someone is unable to activate the new JSON mock support?

panther_analysis_tool/main.py
```diff
@@ -1529,6 +1573,12 @@ def run() -> None:
         RULE_SCHEMA._ignore_extra_keys = True  # pylint: disable=protected-access
         POLICY_SCHEMA._ignore_extra_keys = True  # pylint: disable=protected-access
 
+    # TODO(wey-chiang): Remove in v1.0.0
+    warnings.warn(
+        "Legacy JSON Mocking will be deprecated in v1.0.0 requiring changes to unit tests for"
```
georgepsarakis, line 1578: Maybe:
```suggestion
        "Plain string return values for Mocks will be deprecated in v1.0.0, in favor of full JSON object support, requiring changes to unit tests for"
```

panther_analysis_tool/rule.py
```diff
@@ -133,6 +139,18 @@ def __init__(self, config: Mapping):
         self.detection_id = config["id"]
         self.detection_version = config["versionId"]
 
+        # backwards compatible for legacy mocking
+        self._setup_exception = None
+        if "lastUpdatedVersion" in config:
+            try:
+                detection_panther_version = semver.VersionInfo.parse(config["lastUpdatedVersion"])
+            except Exception as err:  # pylint: disable=broad-except
+                self._setup_exception = err
```
georgepsarakis, line 148: Hmm, this error seems to be Panther's responsibility though, so we should get notified for it, shouldn't we? If you store it as a setup exception it translates to the module not being able to be loaded or something relating to detection code.

panther_analysis_tool/schemas.py
georgepsarakis: Just wondering, why couldn't we provide the hard-coded `LEGACY_MOCKING_VERSION` if the user chooses to use JSON mocks? Otherwise, just pass something like '0.0.0` to fall back?

The `LastUpdatedVersion` field in YAML detection metadata seems as not something that users can easily maintain, it is information set by the backend. In fact, it could potentially create some confusion during the upload, as you would have this field as part of input.

panther_analysis_tool/main.py
wey-chiang: Doh, brain fart -- good catch

panther_analysis_tool/rule.py
wey-chiang: Yeah, I think you're right -- I'll remove the try-except block so the exception is raised

panther_analysis_tool/schemas.py
wey-chiang: > The LastUpdatedVersion field in YAML detection metadata seems as not something that users can easily maintain, it is information set by the backend. In fact, it could potentially create some confusion during the upload, as you would have this field as part of input.

@georgepsarakis That's a good point! So would we have something like a toggle in the UI that allows a user select whether they want to use the JSON mocks?

darwayne:
This has been open forever ... closing out. We can always reference the closed PR if needed

PR #149, opened 2021-10-30
expand regex for allowed Custom schema names

joemiller:
### Background

We have a schema named `Custom.AWS.EKS.Audit` that cannot be used with the PAT test tool. The unit tests will throw an error:

```
[ERROR]: [('./kubernetes_rules/kubectl_exec.yml', SchemaError("Key 'LogTypes' error: LOG_TYPE_REGEX does not match 'Custom.AWS.EKS.Audit'"))]
```

Due to this regex: https://github.com/panther-labs/panther_analysis_tool/blob/fe7fc735b6d62dc585e47349ff58a3dda30961ed/panther_analysis_tool/schemas.py#L54

The rest of the Panther tooling allowed for a schema with this naming scheme to be created and there appears to be precedence for schema names with multiple `.`'s: `Microsoft365\.Audit\.SharePoint`, `Microsoft365\.Audit\.Exchange`.

### Changes

* expand the allowed regex for custom schemas in the test tool to support schema names with multiple separators.

### Testing

* Write a test case .yml file where the `LogTypes` list contains a custom schema name with multiple `.` separators, and then run `panther_analysis_tool test` against it.

### DIFF
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -51,7 +51,7 @@ LOG_TYPE_REGEX = Regex(
     r"Osquery\.Snapshot|Osquery\.Status|OSSEC\.EventInfo|Salesforce\.Login|Salesforce\."
     r"LoginAs|Salesforce\.Logout|Salesforce\.URI|Slack\.AccessLogs|Slack\.AuditLogs|"
     r"Slack\.IntegrationLogs|Sophos\.Central|Suricata\.Anomaly|Suricata\.DNS|Syslog\."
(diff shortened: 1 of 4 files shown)


nhakmiller:
@joemiller I think this should be good (@lindsey-w or @wey-chiang can verify I hope), small thing though we just updated this repo to also have a CLA like `panther-analysis`. So if you could sign that [here](https://cla-assistant.io/panther-labs/panther_analysis_tool) I think we should be good to merge, thanks for the contribution!

joemiller:
@nhakmiller Signed. I think. lmk if I missed something.
thanks!

PR #150, opened 2021-11-04
set suppressions

lindsey-w:
### Background

Suppressions are not correctly being set for detections.

### Changes

* Actually set the suppressions in the detection

### Testing

* `make test`


panther_analysis_tool/rule.py
```diff
@@ -162,6 +162,12 @@ def __init__(self, config: Mapping):
                 values.sort()
             self.detection_reports = config["reports"]
 
+        if "suppressions" in config:
```
georgepsarakis, line 165: Not a problem for sure, it seems that just a more compact version could also work here: 

```python
self.detection_suppressions = config.get("suppressions", [])
```

georgepsarakis (approved):

bseb (approved):

PR #151, opened 2021-11-10
Convert to use Pipenv

bseb:
### Background
Convert to using Pipenv to align with Panther-Analysis

### Changes
- Update Pipfile
- Update Makefile
- Update ci config
-  ~~Remove requirements.txt~~
requirements.txt is still used by parts the CI config

- Added a helper script to update version numbers quickly

### Testing
CI/CD builds, manual testing


Makefile
```diff
@@ -1,32 +1,30 @@
 packages = panther_analysis_tool
 
-ci: lint unit integration
+ci:
+	pipenv run $(MAKE) lint unit integration
```
lindsey-w, line 4: does this end up running `pipenv run make lint` -> `pipenv run pipenv run mypy...`? Curious if we should remove that nesting

bseb: Good catch. That Pipenv run doesn't need to be there. Removed.

bseb: Interestingly CI fails without this, I'm not sure why yet

bseb: Resolved. Found a pip install that was not running in the pipenv environment. Corrected and now works as expected

lindsey-w (approved):

PR #152, opened 2021-11-12
allow OnePassword log type regex

nskobov:
### Background

v1.25.0 introduced OnePassword log types, so I added them to the allowed regex validation so `panther_analysis_tool test` can pass for rules that use those log types.

### Changes

* updated the `LOG_TYPE_REGEX` regex in `schema.py`

### Testing

* `make test`


lindsey-w:
I went ahead and included these with the [PR](https://github.com/panther-labs/panther_analysis_tool/pull/155) for adding zoom logtypes too :)

PR #153, opened 2021-11-17
Add Lookup Table upload support to PAT

matthewcummings:
### Background

It will be helpful for PAT users to be able to perform basic CRUD operations with Lookup Tables.

### Changes

* Added a new lookup_table.py file for the core LUT functionality
* A CLI entrypoint in main.py
* Tests and samples

### Testing

* Unit tests

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -46,6 +46,7 @@ from uuid import uuid4
 
 import botocore
 import requests
+import schema
 from dynaconf import Dynaconf, Validator
 from ruamel.yaml import YAML, SafeConstructor, constructor
 from ruamel.yaml import parser as YAMLParser
@@ -63,6 +64,7 @@ from schema import (
 from panther_analysis_tool.data_model import DataModel
 from panther_analysis_tool.destination import FakeDestination
 from panther_analysis_tool.enriched_event import PantherEvent
+from panther_analysis_tool.lookup_tables import lookup_table
 from panther_analysis_tool.exceptions import UnknownDestinationError
 from panther_analysis_tool.log_schemas import user_defined
 from panther_analysis_tool.policy import TYPE_POLICY, Policy
@@ -70,6 +72,7 @@ from panther_analysis_tool.rule import Detection, Rule
 from panther_analysis_tool.schemas import (
     DATA_MODEL_SCHEMA,
     GLOBAL_SCHEMA,
+    LOOKUP_TABLE_SCHEMA,
     PACK_SCHEMA,
     POLICY_SCHEMA,
     RULE_SCHEMA,
@@ -98,11 +101,12 @@ RULES_PATH_PATTERN = "*rules*"
 DATAMODEL = "datamodel"
 DETECTION = "detection"
 GLOBAL = "global"
+LOOKUP_TABLE = "lookup_table"
 PACK = "pack"
 POLICY = "policy"
 QUERY = "scheduled_query"
-SCHEDULED_RULE = "scheduled_rule"
 RULE = "rule"
+SCHEDULED_RULE = "scheduled_rule"
 
 RESERVED_FUNCTIONS = (
     "alert_context",
@@ -120,6 +124,7 @@ VALID_SEVERITIES = ["INFO", "LOW", "MEDIUM", "HIGH", "CRITICAL"]
 SCHEMAS: Dict[str, Schema] = {
     DATAMODEL: DATA_MODEL_SCHEMA,
     GLOBAL: GLOBAL_SCHEMA,
+    LOOKUP_TABLE: LOOKUP_TABLE,
     PACK: PACK_SCHEMA,
     POLICY: POLICY_SCHEMA,
     QUERY: SCHEDULED_QUERY_SCHEMA,
@@ -254,7 +259,7 @@ def load_analysis_specs(
                         try:
                             yield spec_filename, relative_path, yaml.load(spec_file_obj), None
                         except (YAMLParser.ParserError, YAMLScanner.ScannerError) as err:
-                            # recreate the yaml object and yeild the error
+                            # recreate the yaml object and yield the error
                             yaml = YAML(typ="safe")
                             yield spec_filename, relative_path, None, err
                 if fnmatch(filename, "*.json"):
@@ -486,6 +491,81 @@ def delete_analysis(args: argparse.Namespace) -> Tuple[int, str]:
     logging.info("Detection(s) %s have been deleted.", " ".join(analysis_id_list))
 
     return 0, ""
+
+
+def parse_lookup_table(args: argparse.Namespace) -> dict:
+    """Validates and parses a Lookup Table spec file
+
+    Returns a dict representing the Lookup Table
+
+    Args:
+        args: The populated Argparse namespace with parsed command-line arguments.
+
+    Returns:
+        A dict representing the Lookup Table, empty when parsing fails
+    """
+
+    logging.info("Parsing the Lookup Table spec defined in %s", args.path)
+    with open(args.path, "r") as input_file:
+        try:
+            yaml = YAML(typ="safe")
+            lookup_spec = yaml.load(input_file)
+            logging.info("Successfully parse the Lookup Table file %s", args.path)
+        except (YAMLParser.ParserError, YAMLScanner.ScannerError) as err:
+            logging.error("Failed to parse the Lookup Table spec file %s", input_file)
+            logging.error(err)
+            return {}
+        try:
+            LOOKUP_TABLE_SCHEMA.validate(lookup_spec)
+            logging.info("Successfully validated the Lookup Table file %s", args.path)
+        except (schema.SchemaError, schema.SchemaMissingKeyError, schema.SchemaWrongKeyError,
+                schema.SchemaForbiddenKeyError, schema.SchemaUnexpectedTypeError,
+                schema.SchemaOnlyOneAllowedError) as err:
+            logging.error("Invalid schema in the Lookup Table spec file %s", input_file)
+            logging.error(err)
+            return {}
+    return lookup_spec
+
+
+def test_lookup_table(args: argparse.Namespace) -> Tuple[int, str]:
+    """Validates a Lookup Table spec file
+
+    Returns 1 if the validation fails
+
+    Args:
+        args: The populated Argparse namespace with parsed command-line arguments.
+
+    Returns:
+        A tuple of return code and and empty string (to satisfy calling conventions)
+    """
+
+    logging.info("Validating the Lookup Table spec defined in %s", args.path)
+    lookup_spec = parse_lookup_table(args)
+    if not lookup_spec:
+        return 1,""
+    return 0,""
+
+
+def update_lookup_table(args: argparse.Namespace) -> Tuple[int, str]:
+    """Updates or creates a Lookup Table
+
+    Returns 1 if the update/create fails.
+
+    Args:
+        args: The populated Argparse namespace with parsed command-line arguments.
+
+    Returns:
+        A tuple of return code and an empty string (to satisfy calling conventions)
+    """
+
+    logging.info("Updating the Lookup Table spec defined in %s", args.path)
+    lookup_spec = parse_lookup_table(args)
+    if not lookup_spec:
+        return 1, ""
+    spec_dir = os.path.dirname(args.path)
+    lut = lookup_table.LookupTable(lookup_spec, spec_dir, args.aws_profile)
+    return lut.update(), ""
+>>>>>>> 8a564af (Add Lookup Table upload support to PAT)
 
 
 def update_schemas(args: argparse.Namespace) -> Tuple[int, str]:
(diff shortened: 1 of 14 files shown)


panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,283 @@
...
+import json
+import logging
+from urllib import parse
+
+import boto3
+
+from panther_analysis_tool.util import get_client
+
+LOOKUP_LAMBDA_NAME = "panther-lookup-tables-api"
+LOG_PROCESSOR_LAMBDA_NAME = "panther-log-processor"
```
rleighton, line 31: I would add a comment as to why you need this (to get input bucket).

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,283 @@
...
+            logging.warning("\t[ERROR] Error fetching the input bucket name:\n%s", err)
+            return 1
+        if self.exists():
+            logging.info("Found the Lookup Table %s, updating it now", self.name)
+            return self.update()
+        logging.info("The Lookup Table %s does not exist, creating it now", self.name)
+        return self.create()
+
+
+    # check to see if a LUT with this name already exists (names are unique)
```
rleighton, line 76: You could reduce this code by adding an endpoint to the LUTs API called `createOrUpdate` that implements this logic. The Go code would be shorter also.

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,283 @@
...
+        else:
+            logging.info("The local data file does not match S3 and needs to be uploaded.")
+            if self.upload_data():
+                return 1
+            if self.upload_from_s3():
+                return 1
+        return 0
+
+
+    def get_input_bucket(self) -> str:
```
rleighton, line 177: I suggest you save the value in a class member so the lambda only gets called once.

tests/unit/panther_analysis_tool/test_lookup_tables.py
```diff
@@ -0,0 +1,18 @@
+"""
```
rleighton, line 1: Do you want this empty file?

matthewcummings: It's a placeholder, I still need to add tests specifically for lookup tables. Need to get the existing custom schema tests working again first.

panther_analysis_tool/lookup_table.py
matthewcummings: Yeah, was going to set it via @property but it can raise an exception, which feels slightly odd, but maybe it's still Pythonic.

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,283 @@
...
+        logging.info("Fetching the input bucket name")
+        response = self.client.get_function(
```
georgepsarakis, line 179: I'm not sure if there is need for granting additional permissions to get the function configuration from another Lambda. Alternatively you can create an endpoint in the lookup-tables-api Lambda that just returns the information that `local_file_matches` needs. In general, there is benefit in keeping the client-side simpler (you don't depend so much on version updates for example), so letting the backend handle more complexity can give you more flexibility in the future.

panther_analysis_tool/lookup_table.py
matthewcummings: OK, will move ahead on that front.

panther_analysis_tool/lookup_table.py
matthewcummings: @georgepsarakis yes, good point. However, I could also add the bucket name as an environment variable to the lookup table lambda which is what I was going to do at first but I felt like that would be slightly misleading as the Lambda itself does not need permissions to this bucket. In fact, one of the goals of this PR is to provide an alternate mechanism for uploading files to a lookup table so that we can work around Lambda's payload limit.

But 'saying this out loud', I feel like it's not so bad and is lighter weight to just give the lookup lambda the input bucket name in its configuration (env variables).

rleighton: Actually, just call this endpoint (https://github.com/panther-labs/panther-enterprise/blob/master/api/lambda/logtypes/models_gen.go#L30)  on the logTypes api (PAT should already have permission) that returns an "upload" s3 path. @shaun-sweet is enhancing the LUTs api to be able to handle these.

matthewcummings: OK, to clarify you're recommending that I use this new endpoint to pull the name of which bucket? The one used by the log processor or a different one? Also, this would require that I wait until @shaun-sweet has merged his changes to master.

matthewcummings: @rleighton

panther_analysis_tool/lookup_table.py
matthewcummings: done

panther_analysis_tool/lookup_table.py
matthewcummings: OK, this code will be going away once I package-ize the s3 signing code.

panther_analysis_tool/lookup_table.py
matthewcummings: going away

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,211 @@
...
+        self.spec = spec
+        self.spec_dir = spec_dir
+        self.aws_profile = aws_profile
+        self.client = get_client(self.aws_profile, "lambda")
+        self.lookup_id = None
+        self.input_bucket = ""
+        logging.info("Creating/updating the lookup table %s", self.name)
+
+
+    def create_or_update(self) -> int:
```
darwayne, line 53: `[minor]` what do you think about naming this upsert instead?

darwayne (dismissed):
LGTM

panther_analysis_tool/lookup_table.py
matthewcummings: @darwayne There's another file which has an `update_or_create` method so this felt in line with that style. Doesn't matter too much to me either way. Doesn't really feel like an insert/update/upsert since we're not talking to database directly.

But your team owns this repo so tell me what you want. :)

panther_analysis_tool/main.py
```diff
@@ -288,7 +293,7 @@ def _convert_keys_to_lowercase(mapping: Dict[str, Any]) -> Dict[str, Any]:
 
     Returns:
         A new dictionary with each key converted with str.lower()
-    """
+p    """
```
lindsey-w, line 296:
```suggestion
    """
```

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,171 @@
...
+        An integer return code
+        """
+        logging.info("Creating or updating the lookup table specified in %s", self.spec["dataFile"])
+
+        # First upload the data file to S3
+        upload_url = self.upload_data()
+        if not upload_url:
+            return 1
+
+        request_payload = json.dumps(
```
lindsey-w, line 66: [optional] this would be easier to read if you move the `json.dumps()` into the invoke statement and made this a dictionary with a single field per line

tests/unit/panther_analysis_tool/test_lookup_tables.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+from unittest import TestCase
+
+from panther_analysis_tool import main as pat
+
+FIXTURES_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../', 'fixtures'))
+LUTS_FIXTURES_PATH = os.path.join(FIXTURES_PATH, 'lookup-tables')
+
+class TestLookupTable(TestCase):  # pylint: disable=too-many-public-methods
+    def test_load_policy_specs_from_folder(self):
```
lindsey-w, line 30:
```suggestion
    def test_load_invalid_specs_from_folder(self):
```

lindsey-w (commented):
Looks good! Might be nice to include additional unit tests for the LookupTable methods?

panther_analysis_tool/lookup_table.py
matthewcummings: I did that so that I could add the logging.debug() line. Can remove if you don't think that helps. @lindsey-w

matthewcummings: @lindsey-w I considered adding more tests but as unit tests, we can't really test much with this code other than that it parses responses output correctly. Also, I have integration tests in the backend which cover the API functionality.

It would be nice to have a framework in place that ran PAT tests against an actual backend but it would take some effort to build that out.

lindsey-w (dismissed):

panther_analysis_tool/main.py
```diff
@@ -398,6 +403,80 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+            yaml = YAML(typ="safe")
+            lookup_spec = yaml.load(input_file)
+            logging.info("Successfully parse the Lookup Table file %s", args.path)
+        except (YAMLParser.ParserError, YAMLScanner.ScannerError) as err:
+            logging.error("Failed to parse the Lookup Table spec file %s", input_file)
+            logging.error(err)
+            return {}
+        try:
+            LOOKUP_TABLE_SCHEMA.validate(lookup_spec)
+            logging.info("Successfully validated the Lookup Table file %s", args.path)
```
georgepsarakis, line 430: Maybe use debug level?

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,174 @@
...
+
+        response_str = response["Payload"].read().decode("utf-8")
+        response_payload = json.loads(response_str)
+        logging.debug(response_payload)
+        status_code = response["StatusCode"]
+        error_msg = response_payload.get("errorMessage", None)
+
+        if status_code != 200 or error_msg:
+            logging.error(
+                "Failed to upload the S3 file %s to the %s Lookup Table\n\tstatus code: %s \
```
georgepsarakis, line 166: Same note about logging presigned URLs as above, in case you find it important.

georgepsarakis (commented):
Nice work!

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,174 @@
+"""
```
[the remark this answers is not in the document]
matthewcummings: I could but my team doesn't "own" this code + it's public + even if we did own this code, I'm really not trying to overhaul everything because my first thought is the several other files in this directory should also be their own packages.

panther_analysis_tool/main.py
matthewcummings: I would say no, I usually lean towards verbosity with tools like PAT which run in CI/CD pipelines since it makes debugging easier and this is not super verbose in my view.

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,174 @@
...
+        response_str = response["Payload"].read().decode("utf-8")
+        response_payload = json.loads(response_str)
+        logging.debug(response_payload)
+
+        self.lookup_id = response_payload["id"]
+        error_msg = response_payload.get("errorMessage", None)
+        status_code = response["StatusCode"]
+        if status_code != 200 or error_msg:
+            logging.warning(
+                "Failed to create/update the %s Lookup Table\n\tstatus code: %s\n\terror: %s",
```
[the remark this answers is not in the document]
matthewcummings: This is copy/pasta from other parts of the codebase. In many cases I've just been keeping with the "style" of this code.

panther_analysis_tool/lookup_table.py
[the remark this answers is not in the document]
matthewcummings: OK, I'll do it. :)

panther_analysis_tool/lookup_table.py
```diff
@@ -0,0 +1,174 @@
...
+        response = self.client.invoke(
+            FunctionName=LOOKUP_LAMBDA_NAME,
+            InvocationType="RequestResponse",
+            Payload=request_payload,
+        )
+        response_str = response["Payload"].read().decode("utf-8")
+        response_payload = json.loads(response_str)
+        logging.debug(response_payload)
+
+        self.lookup_id = response_payload["id"]
```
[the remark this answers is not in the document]
matthewcummings: Lookup tables names are unique.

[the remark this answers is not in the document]
georgepsarakis: Aha thanks for clarifying!

panther_analysis_tool/main.py
georgepsarakis: Sure, makes sense.

panther_analysis_tool/lookup_table.py
darwayne: @matthewcummings got it. Anytime I see a function that is only responsible for creating or updating I recommend naming it upsert since it is concise an encapsulates both create and update. create is synonymous to insert. So upsert doesn't necessarily imply what your underlying storage mechanism is.

darwayne (dismissed):

panther_analysis_tool/main.py
```diff
@@ -63,13 +64,15 @@
 from panther_analysis_tool.data_model import DataModel
 from panther_analysis_tool.destination import FakeDestination
 from panther_analysis_tool.enriched_event import PantherEvent
+from panther_analysis_tool.lookup_tables import lookup_table
```
georgepsarakis, line 67: Reminder to include the package in `setup.py` (see [here](https://github.com/panther-labs/panther_analysis_tool/blob/199c2d294e25b9ef332045996544974ab970ed87/setup.py#L24)) so that it is included in the PyPI release.

darwayne (approved):

darwayne (approved):

PR #154, opened 2021-11-17
FIX: Filtering Boolean Fields

wey-chiang:
### References

Closes https://app.asana.com/0/1201300875909824/1201247825599117/f

### Background

The logic for parsing filters breaks for the `Enabled` field i.e.  `--filter Enabled=true` due to improper conversion from a string to boolean. Since the `Enabled` field loaded from the YAML file returns a boolean, this improper conversion caused this filter to not work.

Note: currently the `Enabled` field is the only boolean field, besides Tests:ExpectedResult which users cannot filter on.

### Changes

* Added logic to handle boolean fields
* Added unit tests

### Testing

* See Attached Screenshots

#### Unit Test Duplicating Bug

Note that the `Example.DisabledRule` is evaluated here when it shouldn't be with the `Enabled!=false` flag.

image

#### Unit Test w/ Fix

Note that the `Example.DisabledRule` is properly filtered now.

image


tests/fixtures/detections/disabled_rule/example_rule.yml
```diff
@@ -0,0 +1,28 @@
...
+Runbook: >
+  Find out who disabled MFA on the account.
+Reference: https://www.link-to-info.io
+Tests:
+  -
+    Name: Example Test
+    ExpectedResult: true
+    Log:
+      foo: bar
+
```
bseb, line 28: Nit: Looks like a minor formatting issue here. Does the new line have trailing whitespace maybe?

bseb (approved):
Nice! 

Left one comment but not a blocker for merging

lindsey-w (approved):
Nice! 🥳

PR #157, opened 2021-12-16
Add Atlassian to log regex

bseb:
### Background
New Log Type for Atlassian

### Changes
Added Atlassian to Log Regex

### Testing

Unit tests

### DIFF
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -29,7 +29,7 @@ RESOURCE_TYPE_REGEX = Regex(
     r"\.Instance|Redshift\.Cluster|S3\.Bucket|WAF\.Regional\.WebACL|WAF\.WebACL)$"
 )
 LOG_TYPE_REGEX = Regex(
-    r"^(Apache\.AccessCombined|Apache\.AccessCommon|AWS\.ALB|AWS\.AuroraMySQLAudit|AWS"
+    r"^(Apache\.AccessCombined|Apache\.AccessCommon|Atlassian\.Audit|AWS\.ALB|AWS\.AuroraMySQLAudit|AWS"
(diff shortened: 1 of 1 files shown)


panther_analysis_tool/schemas.py
```diff
@@ -29,9 +29,10 @@
     r"\.Instance|Redshift\.Cluster|S3\.Bucket|WAF\.Regional\.WebACL|WAF\.WebACL)$"
 )
 LOG_TYPE_REGEX = Regex(
-    r"^(Apache\.AccessCombined|Apache\.AccessCommon|AWS\.ALB|AWS\.AuroraMySQLAudit|AWS"
```
bseb, line 32: Maybe we want to bump or ignore line length in this stanza? This is kinda painful to shuffle when trying to maintain alphabetical order

darwayne (approved):

panther_analysis_tool/schemas.py
```diff
@@ -29,9 +29,10 @@
...
-    r"\.CloudTrail|AWS\.CloudTrailDigest|AWS\.CloudTrailInsight|AWS\.CloudWatchEvents|AWS"
-    r"\.GuardDuty|AWS\.S3ServerAccess|AWS\.VPCDns|AWS\.VPCFlow|AWS\.WAFWebACL|Box\.Event"
+    r"^(Apache\.AccessCombined|Apache\.AccessCommon|Atlassian\.Audit|AWS\.ALB"
```
nmartin413, line 32: should the `(` be unterminated?

bseb: It actually terminates on like 57. This is a wild regex.

bseb: It actually terminates on line 57.

nmartin413 (approved):

panther_analysis_tool/schemas.py
lindsey-w: agreed, this is super annoying and hard to maintain

PR #158, opened 2022-01-03
adds Dar and Nic to code owners

darwayne:
### Background

Adds more members of the team to simplify merges


nhakmiller (approved):
Thanks for putting this in. No idea why it's failing CI, looks like some kind of dependency mess up maybe?

bseb (approved):

PR #162, opened 2022-01-20
Add Delete Function and bump Version to 0.11.0

bseb:
### Background
Relates to https://app.asana.com/0/1201638421895884/1200545021201180/f
Bump version to 0.11.0
Adds ability to delete detections from PAT

### Changes
Added Delete function and argument parser

### Testing
Manual testing of functionality via PAT
CI/Lint/Unit tests


panther_analysis_tool/main.py
```diff
@@ -398,6 +399,90 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+        for detection in json_output["detections"]:
+            analysis_found[detection["id"]] = detection
+        # Build a list of analysis we've confirmed are in Panther
+        for analysis in analysis_id_list:
+            if analysis not in analysis_found:
+                logging.info("%s was not found, skipping...", analysis)
+            else:
+                analysis_confirmed.append(analysis)
+        return analysis_confirmed
```
lindsey-w, line 429: I think you could potentially get rid of this list and just return the keys from the dict 
```suggestion
        return analysis_found.keys()
```

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,90 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+    # Validate the Detection that we are deleting exists in Panther
+    # If nothing was found bail out, otherwise prepare to delete what has been confirmed
+    analysis_confirmed = confirm_analysis_exists(args, analysis_id_list)
+
+    if len(analysis_confirmed) == 0:
+        logging.error("No matching analysis found, exiting")
+        return 1, ""
+
+    analysis_id_list = analysis_confirmed
```
lindsey-w, line 448: wondering if you could just set the return to analysis_id_list? 

```suggestion
    payload: dict = {"deleteDetections": {"entries": []}}

    # Validate the Detection that we are deleting exists in Panther
    # If nothing was found bail out, otherwise prepare to delete what has been confirmed
    analysis_id_list = confirm_analysis_exists(args, args.analysis_id)

    if len(analysis_id_list) == 0:
        logging.error("No matching analysis found, exiting")
        return 1, ""
```

lindsey-w:
Nice work!

panther_analysis_tool/main.py
bseb: Yup, good idea. Apparently it has to be `list[analysis_found.keys()` because the `.keys()` function returns `dict_keys` object. TIL

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,95 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+    validation = client.invoke(
+        FunctionName="panther-analysis-api",
+        InvocationType="RequestResponse",
+        LogType="None",
+        Payload=json.dumps(validation_payload),
+    )
+
+    validation_string = json.loads(validation["Payload"].read().decode("utf-8"))["body"]
+    # To handle mocks in testing, converting to string does not result in a processable string
+    if isinstance(validation_string, dict):
```
bseb, line 415: I really am not a fan of this, but in testing the validation string is returned as a dict. Type casting it to a string results in a string that can not be parsed by the json module because of a the entities are not in double quotes, attempting to replace the single quotes with double quotes results in invalid json.  

The machinations of trying to manipulate it back into a processesable string where getting pretty ugly ( and still not working), and this seemed like a more simple solution rather than deep diving on what was a happening in the mock. I am absolutely open to other ideas if anyone has them.

cc: @wey-chiang  This is what I wanted to talk about at 3

panther_analysis_tool/main.py
bseb: Done and done

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,94 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+    api_response = json.loads(validation["Payload"].read().decode("utf-8"))
+    json_response = api_response["body"]
+
+    # The mock will result in a dict, the actual api returns a string.
+    # The following block deals with that deviation
+    if isinstance(json_response, dict):
+        json_output = json_response
+    # Normal Execution will hit the else clause
+    else:
+        json_output = json.loads(json_response)
```
bseb, line 422: As mentioned in the comment, the api returns a string as the value for `body` where as the mock returns a dict. I tried a lot of iterations of this to make handling the deviation as clean as possible and this is where I landed ( with a significant amount of help from @wey-chiang ). All the typecasting and string foo was quite a bit uglier.

All that said, if there is a better way I'm very open to it. Tagging the Python gurus @lindsey-w @georgepsarakis  @wey-chiang @nhakmiller

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,94 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+    if len(analysis_id_list) == 0:
+        logging.error("No matching analysis found, exiting")
+        return 1, ""
+
+    # Get user confirmation to delete
+    analysis_id_string = " ".join(analysis_id_list)
+    logging.warning("You are about to delete detections %s", analysis_id_string)
+    confirm = input("Continue? (y/n)")
+
+    if confirm.lower() == "y":
```
georgepsarakis, line 459: An early return on the cancellation selection would likely be a bit easier to follow. Essentially this translates to avoiding large nested code regions in `if` statements.

```python
if confirm.lower() != "y":
  print("Cancelled")
  return 0, ""
```

tests/unit/panther_analysis_tool/test_main.py
```diff
@@ -471,5 +471,26 @@ def test_available_destination_names_valid_name_returned(self):
...
+        from unittest import mock
+        """Validation function only returns analysis that exist in Panther"""
+        requested_deletion = ['Rule.Does.Exist', 'Rule.Doesnt.Exist']
+        args = pat.setup_parser().parse_args(f'delete '
+                                             f'--analysis-id '
+                                             f'{requested_deletion}'.split())
+
+        rv = '{"body":{"paging": {"thisPage": 1, "totalPages": 1, "totalItems": 1}, "detections": [{"complianceStatus": "", "resourceTypes": [], "suppressions": [], "dedupPeriodMinutes": 0, "logTypes": [], "scheduledQueries": [], "summaryAttributes": [], "threshold": 0, "analysisType": "", "body": "", "createdAt": "0001-01-01T00:00:00Z", "createdBy": "", "description": "", "displayName": "", "enabled": false, "id": "Rule.Does.Exist", "lastModified": "0001-01-01T00:00:00Z", "lastModifiedBy": "", "outputIds": [], "packIds": [], "reference": "", "reports": {}, "runbook": "", "severity": "", "tags": [], "tests": [], "versionId": "", "managed": false, "parentId": ""}]}}'
+
+        class MockObject():
```
georgepsarakis, line 486: I don't think you need this class. You can set `rv` as a `bytes` multiline literal so that it is also easier to read:

```python
rv = b'''
{
  "body": ...
}
'''
```

And then use a `BytesIO` instance as the return value for the `invoke` method.

```python
from io import BytesIO

invoke_mock.invoke.return_value = BytesIO(rv)
```



I believe this will also resolve [your question above](https://github.com/panther-labs/panther_analysis_tool/pull/162/files#r793115177).

panther_analysis_tool/main.py
bseb: Done, good suggestion

tests/unit/panther_analysis_tool/test_main.py
bseb: Thank you @georgepsarakis , that plus a little bit of manipulation of the payload created a correct mock. Much appreciated

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,88 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+    analysis_id_list = confirm_analysis_exists(args, analysis_id_list)
+    if len(analysis_id_list) == 0:
+        logging.error("No matching analysis found, exiting")
+        return 1, ""
+
+    # Get user confirmation to delete
+    analysis_id_string = " ".join(analysis_id_list)
+    logging.warning("You are about to delete detections %s", analysis_id_string)
+    confirm = input("Continue? (y/n)")
```
wey-chiang, line 451: Nit: add a space after the closing parentheses 
```suggestion
    confirm = input("Continue? (y/n) ")
```

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,88 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+    if response_payload.get("statusCode") != 200:
+        logging.warning(
+            "Failed to delete analysis.\n\tstatus code: %s\n\terror message: %s",
+            response_payload.get("statusCode", 0),
+            response_payload.get("errorMessage", response_payload.get("body")),
+        )
+        return 1, ""
+
+    logging.info("Detection has been deleted.")
```
wey-chiang, line 479: It'd be helpful to list the detections that were deleted here or at least the number of detections deleted

wey-chiang (dismissed):
LGTM

panther_analysis_tool/main.py
bseb: Added

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,88 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+        LogType="None",
+        Payload=json.dumps(validation_payload),
+    )
+
+    api_response = json.loads(validation["Payload"].read().decode("utf-8"))
+    json_response = api_response["body"]
+
+    json_output = json.loads(json_response)
+
+    if len(analysis_id_list) != json_output["paging"]["totalItems"]:
```
darwayne, line 418: if lambda throws an error will this cause an exception here?

panther_analysis_tool/main.py
```diff
@@ -398,6 +399,88 @@ def upload_analysis(args: argparse.Namespace) -> Tuple[int, str]:
...
+        FunctionName="panther-analysis-api",
+        InvocationType="RequestResponse",
+        LogType="None",
+        Payload=json.dumps(payload),
+    )
+
+    response_str = response["Payload"].read().decode("utf-8")
+    response_payload = json.loads(response_str)
+
+    if response_payload.get("statusCode") != 200:
```
darwayne, line 471: this is nice logic, feel like you might want to also use it in the confirm func as well

darwayne (dismissed):
looks good for the most part, think we can do a better job of error handling but it isn't a deal breaker for me

panther_analysis_tool/main.py
bseb: You are right, I didn't consider that when I split the function up. Added error handling, the logic differs slightly because  this is not the same API end point and responds differently

panther_analysis_tool/main.py
bseb: Added some error handling

darwayne (approved):

wey-chiang (approved):

PR #170, opened 2022-03-30
LUTs Pack updates

matthewcummings:
### Background

Update PAT to handle the new LUT schema. Remove the update lookup table functionality as we'll be managing LUTs via Packs going forward. 

### Changes

* Removed `update_lookup_table`
* Updated the LUT schema
* Updated relevant tests

### Testing

* Unit tests
* Manual usage with internal security Panther deployments


panther_analysis_tool/main.py
```diff
@@ -230,6 +231,7 @@ def load_analysis_specs(
                         fnmatch(relative_path, path_pattern)
                         for path_pattern in (
                             DATA_MODEL_PATH_PATTERN,
```
rleighton, line 233: Do we need a LUTS patch where customers can put THEIR LUTs so they would get packaged? Perhaps we change `GREYNOISE_LUTS_PATH_PATTERN` to `LUTS_PATH_PATTERN,` and nest greynoise under a `lookup_tables` folder?

matthewcummings: good call

rleighton (approved):
LGTM but you should wait on the detections team to merge

darwayne (approved):

PR #174, opened 2022-04-06
Fix 'Fuzzy Search' Bug in PAT

bseb:
### Background
This Corrects the bug of returning items that were an approximate match for the query

### Changes

* Change analysis validation function to return only items that were validated AND passed on the command line

### Testing

* Unit testing, also ran through manual testing scenarios of deleting partial matches to ensure they were not returning anything.


bseb:
@lindsey-w  @darwayne  I'd love if you have me a second set of eyes here. Manual testing of this looks good to me

darwayne (approved):

panther_analysis_tool/main.py
```diff
@@ -475,18 +474,22 @@ def confirm_analysis_exists(args: argparse.Namespace, analysis_id_list: list) ->
     json_response = api_response["body"]
 
     json_output = json.loads(json_response)
-
     if len(analysis_id_list) != json_output["paging"]["totalItems"]:
-        analysis_found = {}
+
+        analysis_response = {}
```
georgepsarakis, line 479: Just a suggestion, feel free to ignore if you don't find it easier to maintain, I think you can achieve the same with sets:

```python3
existing_ids = set([detection["id"] for detection in json_output["detections"]])
given_ids = set(analysis_id_list)
diff = given_ids.difference(existing_ids)
if diff:
    for detection_id in diff:
         # log 
    return list(existing_ids.intersection(given_ids))
return analysis_id_list
```

bseb:
This validation check is no longer behind an if statement. When re-testing after Georges suggestion I was able to make it return the wrong rules. Now everyone gets validated.

darwayne (approved):

lindsey-w (approved):

PR #177, opened 2022-04-25
add simple pack validation

lindsey-w:
### Background

Relates to: https://app.asana.com/0/1200821831799747/1202171285834906/f

### Changes

* First step towards adding some additional validation to packs: check that all the ids specified in the file actually exist
    * another future improvement might be to check that all the required dependencies for the detections in the pack have been included.

### Testing

* `pat test --path ./path-to-panther-analysis`

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -314,7 +314,7 @@ def zip_analysis(args: argparse.Namespace) -> Tuple[int, str]:
         if return_code != 0:
             return return_code, ""
 
-    logging.info("Zipping analysis packs in %s to %s", args.path, args.out)
+    logging.info("Zipping analysis items in %s to %s", args.path, args.out)
     # example: 2019-08-05T18-23-25
     # The colon character is not valid in filenames.
     current_time = datetime.now().isoformat(timespec="seconds").replace(":", "-")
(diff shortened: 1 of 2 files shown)


panther_analysis_tool/main.py
```diff
@@ -1163,6 +1167,39 @@ def setup_run_tests(  # pylint: disable=too-many-locals,too-many-arguments
...
+    for analysis_type in analysis_specs:
+        for analysis_spec_filename, _, analysis_spec in analysis_specs[analysis_type]:
+            analysis_id = (
+                analysis_spec.get("PolicyID")
+                or analysis_spec.get("RuleID")
+                or analysis_spec.get("DataModelID")
+                or analysis_spec.get("GlobalID")
+                or analysis_spec.get("PackID")
+                or analysis_spec.get("QueryName")
+                or analysis_spec["LookupName"]
```
darwayne, line 1184: any reason we aren't using `get` here like all the others?

darwayne (dismissed):

panther_analysis_tool/main.py
lindsey-w: I left the last one as a regular access for the case where we add another analysis type in the future - we would want this to error so we can update the list of possible locations the id can come from

darwayne (approved):

danbiwer (approved):

PR #180, opened 2022-04-28
global helpers from dir

lindsey-w:
### Background

We should be able to import globals form other globals.  This change add globals to a temp directory and add that dir to the path, allowing globals to import other globals.  The directory is cleared before adding to the path/adding files to it.

### Changes

* import globals from directory

### Testing

* `make test`
* `pat ./panther-analysis`


darwayne (approved):

danbiwer (approved):

bseb (approved):
I think this is solid. The directory created seems to be at least some amount of randomly named ( Repeatedly the same on one machine, but not the same across machines? ) and is handled securely based on docs here https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile


LGTM

panther_analysis_tool/main.py
```diff
@@ -97,6 +99,7 @@
 POLICIES_PATH_PATTERN = "*policies*"
 QUERIES_PATH_PATTERN = "*queries*"
 RULES_PATH_PATTERN = "*rules*"
+TMP_HELPER_MODULE_LOCATION = os.path.join(tempfile.gettempdir(), "globals")
```
georgepsarakis, line 102: `globals` may be a bit generic. Consider using a namespace (additional directory), like `panther-pat` or similar.

panther_analysis_tool/main.py
```diff
@@ -1048,22 +1050,39 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
...
-            invalid_specs.append((analysis_spec_filename, load_err))
-            continue
-        sys.modules[analysis_id] = module
-    return invalid_specs
-
+        source = os.path.join(dir_name, analysis_spec["Filename"])
+        destination = os.path.join(TMP_HELPER_MODULE_LOCATION, f"{analysis_id}.py")
+        shutil.copyfile(source, destination)
+        # force reload of the module as necessary
+        if analysis_id in sys.modules:
```
georgepsarakis, line 1074: I think you should be raising an exception here. Couldn't there be a conflict with a standard library module for example? For example, someone could name their helper `json` by mistake.

panther_analysis_tool/main.py
```diff
@@ -1048,22 +1050,39 @@ def test_analysis(args: argparse.Namespace) -> Tuple[int, list]:
...
+            importlib.reload(sys.modules[analysis_id])
+
+def cleanup_global_helpers(global_analysis: List[Any]) -> None:
+    # clear the modules from the modules cache
+    for _, _, analysis_spec in global_analysis:
+        analysis_id = analysis_spec["GlobalID"]
+        if analysis_id in sys.modules:
```
georgepsarakis, line 1081: Maybe you should note that this is for use in tests? Or isn't this the only case?

tests/fixtures/detections/valid_analysis/global_helpers/a_helper.py
```diff
@@ -0,0 +1,4 @@
+from b_helper import b_says_hello
```
georgepsarakis, line 1: 🎉

georgepsarakis (commented):
Nice work!

PR #187, opened 2022-07-05
PAT Public API

nmartin413:
Updates PAT to use either the lambda interface or a new graphql interface.

- refactors the calls into a generic interface
- consolidates the lambda calls
- adds graphql support (and new arguments for API host and token)
- adds a connection check method (Public API only)
- removes unused code


darwayne (commented):
🤩 looking good so far

panther_analysis_tool/backend/public_api_client.py
```diff
@@ -0,0 +1,145 @@
...
+    def delete_detections(self, params: DeleteDetectionsParams) -> BackendResponse:
+        pass
+
+    def _execute(self, request: DocumentNode) -> ExecutionResult:
+        return self._gql_client.execute(request, get_execution_result=True)
+
+
+_API_URL_PATH = "public/graphql"
+_API_DOMAIN_PREFIX = "api"
+_API_TOKEN_HEADER = "X-API-Key"
```
darwayne, line 127: I think you'll need to throw in a lint ignore flag here to appease the ci gods

danbiwer (approved):
Nice, LGTM.

I suppose we will also want to add `putSchema`and `listSchemas`.  I think you actually had those in at some point before I said PAT wasn't using them 😬

kostaspap:
> Nice, LGTM.
> 
> I suppose we will also want to add `putSchema`and `listSchemas`. I think you actually had those in at some point before I said PAT wasn't using them 😬

Yup, we have customer using PAT to manage their custom schemas!

danbiwer (approved):
awesome, huge improvement

danbiwer (approved):

darwayne (approved):

PR #190, opened 2022-07-21
adds public api bulk upload support

darwayne:
### Background

Adds bulk upload support via the api token

### Changes

* Adds real implementation for the public api client around bulk uploads

### DIFF
--- a/panther_analysis_tool/backend/public_api_client.py
+++ b/panther_analysis_tool/backend/public_api_client.py
@@ -17,12 +17,11 @@ You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <https://www.gnu.org/licenses/>.
 """
 
-import os
 import logging
-
-from typing import Dict
-from pathlib import Path
+import os
 from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, Optional, Any
 from urllib.parse import urlparse
 
 from gql import Client as GraphQLClient, gql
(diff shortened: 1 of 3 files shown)


panther_analysis_tool/backend/public_api_client.py
```diff
@@ -120,8 +150,8 @@ def list_managed_schema_updates(self) -> BackendResponse:
     def update_managed_schemas(self, params: UpdateManagedSchemasParams) -> BackendResponse:
         pass
 
-    def _execute(self, request: DocumentNode) -> ExecutionResult:
-        return self._gql_client.execute(request, get_execution_result=True)
+    def _execute(self, request: DocumentNode, variable_values: Optional[Dict[str, Any]] = None, ) -> ExecutionResult:
```
danbiwer, line 153: cool, this is exactly what I did in my branch

panther_analysis_tool/backend/public_api_client.py
```diff
@@ -106,7 +110,33 @@ def check(self) -> BackendCheckResponse:
...
+        query = self._requests.bulk_upload_mutation()
+        upload_params = {"input": {"data": params.encoded_bytes()}}
+        default_stats = dict(total=0, new=0, modified=0)
+        res = res = self._execute(query, variable_values=upload_params)
+
+        if res.errors:
+            for err in res.errors:
+                logging.error(err.message)
+
+            return BackendCheckResponse(
```
nmartin413, line 122: `BackendCheckResponse` was meant to be a special value that only came from `def check()`. We could probably eliminate it and use the generic `BackendResponse[CheckResponse]` format instead 

Might want to extend `BackendResponse` to have an error field?

darwayne: woah good catch

danbiwer (approved):
awesome 💯

PR #196, opened 2022-07-25
update-custom-schemas uses Backend Client

danbiwer:
### Background

This sets up the update-custom-schemas operations to use the backend client so that we can support GraphQL and direct lambda invocations.

### Changes

* Adding lambda and graphql implementations of List/Update schemas for backend client
* Refacter of log schemas updaters to take backend client as dependency
* stripping down mock backend client in favor of using MagicMock

### Testing

* unit tests / manual tests


example_panther_config.yml
```diff
@@ -2,6 +2,12 @@
 # Copy this file into the root of your test repository and name it .panther_settings.yml
 # Note that options in this file override options passed on the command line
 #
+# The API token to use when making changes to Panther
```
darwayne, line 5: ❤️

darwayne (approved):

nmartin413 (dismissed):

darwayne (approved):

PR #199, opened 2022-07-29
read variables from env

danbiwer:
### Background

Env variables were only read if panther_config.yml existed.  Now the env variables are always parsed.

The variable precedence is now environment > panther_config.yml > command line options

### Changes

* Environment variables are always parsed

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -1761,16 +1761,22 @@ def run() -> None:
     if getattr(args, "filter_inverted", None) is None:
         args.filter_inverted = {}
 
+    for key in os.environ:
+        if key.startswith("PANTHER_"):
+            logging.info(
+                "Found Environment Variables prefixed with 'PANTHER'.  NOTE: ENVIRONMENT VARIABLES OVERRIDE COMMAND LINE OPTIONS"
+            )
+            break
     if os.path.exists(CONFIG_FILE):
         logging.info(
             "Found Config File %s . NOTE: SETTINGS IN CONFIG FILE OVERRIDE COMMAND LINE OPTIONS",
             CONFIG_FILE,
         )
-        config_file_settings = setup_dynaconf()
-        dynaconf_argparse_merge(vars(args), config_file_settings)
-        if args.debug:
-            for key, value in vars(args).items():
-                logging.debug(f"{key}={value}")  # pylint: disable=W1203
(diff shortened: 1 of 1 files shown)


lindsey-w (dismissed):

panther_analysis_tool/main.py
```diff
@@ -1761,16 +1761,22 @@ def run() -> None:
     if getattr(args, "filter_inverted", None) is None:
         args.filter_inverted = {}
 
+    for key in os.environ:
+        if key.startswith("PANTHER_"):
+            logging.info(
+                "Found Environment Variables prefixed with 'PANTHER'.  NOTE: ENVIRONMENT VARIABLES OVERRIDE COMMAND LINE OPTIONS"
```
nmartin413, line 1767: can we invert this? i think it would be more expected that cli opts take precedence

nmartin413: aws cli precedence, for example:

image

can fish up more if necessary

nmartin413 (commented):
commented on concerns w/ precedence decision.

panther_analysis_tool/main.py
danbiwer: updated.

Best I can do is command line > environment > panther_config (these last two are set by dynaconf)

panther_analysis_tool/main.py
```diff
@@ -1694,8 +1694,19 @@ def setup_dynaconf() -> Dict[str, Any]:
 def dynaconf_argparse_merge(
         argparse_dict: Dict[str, Any], config_file_settings: Dict[str, Any]
 ) -> None:
+    # Set up another parser w/ no defaults
+    aux_parser = argparse.ArgumentParser(argument_default=argparse.SUPPRESS)
```
danbiwer, line 1698: This feels kinda janky, but this is the only way I could find to get the desired precedence without changing our arg parsing library 

based on [this](https://stackoverflow.com/a/45803037)

nmartin413: i think the way we're getting input needs to be refactored at some point, so as long as this establishes the standard precedence of `file->env->arg` I think it's OK

darwayne (approved):

panther_analysis_tool/main.py
danbiwer: I'm taking this as implied approval

PR #202, opened 2022-08-04
deps update + prep for release

lindsey-w:
### Background

Update dependencies + prep for release

### Changes

* update deps
* prep for next `pat` release

### Testing

* `make test`
* manual testing:
* `python setup.py sdist` -> pip3 install -> `pat upload`


danbiwer:
I can build this successfully and run all commands, but the circle ci build step is failing.  The module 'typing_extensions' isn't referenced anywhere in this repo, and I can run the failing ci command locally with no errors.

Any ideas here?  @michael-cariello @cdzombak @lindsey-w

danbiwer:
Hijacking this PR:

This fixes setup.py so that we don't have missing dependencies

danbiwer (dismissed):

darwayne (approved):

Pipfile
```diff
@@ -8,7 +8,7 @@ bandit = "==1.7.0"
 black = "==20.8b1"
 isort = "==5.8.0"
 mypy = "==0.812"
-panther-core = "==0.0.9"
+panther-core = "==0.0.10"
```
darwayne, line 11: is this right? shouldn't this be 0.1.0?

darwayne (changes requested):
yeah we should make sure version is right: https://pypi.org/project/panther-core/0.2.0/#history

darwayne:
P.S ... just added a 0.2.0 for panther_core so we could go with that .. should have non breaking changes

darwayne:
not sure if 0.0.9 -> 0.1.0 had breaking changes though

Pipfile
danbiwer: currently the detections engine/executor are using 0.0.10.  Is that going to change in this upcoming release?

danbiwer: from release-1.40's requirements.txt
```
panther-core==0.0.10
```

darwayne: ahh I missed that we actually went up to 0.0.10 in the pip .. my bad ... yeah we have some newer versions coming that should probably go out in a separate PR

darwayne (approved):

darwayne:
Ok I was wrong, didn't realize we had a 0.0.10 version .. saw 0.1.0 and thought this was wrong

nmartin413 (approved):

PR #205, opened 2022-08-08
Update Lookup Table schema with a new field

dashaaa:
### Background

Now users can upload lookup table content by specifying it in the FileName field. The DataFile field were a breadcrumbs from previous implementations which is not included here. Supporting of new file extensions is added as well (WIP)

### Changes

* Add a new field to the schema (instead of an obsolete one)

### Testing

* Unit test is edited to match code changes 
* I checked manually that uploading works

### DIFF
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -204,7 +204,7 @@ LOOKUP_TABLE_SCHEMA = Schema(
         "AnalysisType": Or("lookup_table"),
         "LookupName": str,
         "Enabled": bool,
-        Or("DataFile", "Refresh"): Or(
+        Or("FileName", "Refresh"): Or(
             str,
             {
                 "RoleARN": str,
--- a/panther_analysis_tool/schemas.py
+++ b/panther_analysis_tool/schemas.py
@@ -204,7 +204,7 @@ LOOKUP_TABLE_SCHEMA = Schema(
         "AnalysisType": Or("lookup_table"),
         "LookupName": str,
         "Enabled": bool,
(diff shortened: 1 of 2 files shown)


panther_analysis_tool/schemas.py
```diff
@@ -204,7 +204,7 @@
         "AnalysisType": Or("lookup_table"),
         "LookupName": str,
         "Enabled": bool,
-        Or("DataFile", "Refresh"): Or(
+        Or("FileName", "Refresh"): Or(
```
darwayne, line 207: not clear on what the implication are around this. When you're free can we have a quick chat in slack/zoom?

darwayne:
image

darwayne:
as discussed in our offline chat, looks like one more piece that's needed is to make sure PAT zips the file you expect it to

dashaaa:
> 

Changing FileName -> Filename makes PAT to zip file that is mentioned in this field (similarly to .py files), so the last comment is taken care of

dashaaa:
@darwayne do you mind to take another look?

darwayne:
@dashaaa will do, code looks good, have you done any manual testing to verify this works as expected?

dashaaa:
@darwayne Yes! I was able to upload lookup table yml + data. It works when I do it from the directory itself with `panther_analysis_tool --debug upload --path ./`. I was not able to do it from another directory though, but I am not clear on how it is supposed to work. Please, advise if you know.

darwayne:
that sounds like a [known issue](https://app.asana.com/0/1200908948600035/1202486395033678/f)

darwayne (approved):

PR #206, opened 2022-08-12
feat: update panther_core to be able to support unit testing of title function output

edyesed:
### Background

**WARNING: NOT READY FOR MERGE AT THIS TIME**

As a security engineer, it would be helpful to me to be able to unit test the output of the title function. This requires [panther_core PR 23](https://github.com/panther-labs/panther_core/pull/23), and the Pipfile at this time is calling panther_core by hash. 

### Changes

*  Unit test the title function in a detection

### Testing

* test run locally, see attached screencaps
* If the `ExpectedTitle` attribute is set on a TestCase, then we evaluate if the title function's output matches the test condition. The test case fails overall if it does not match. I re-used the exception raising in the test cases to surface the difference between expected and found. 
* If the `ExpectedTitle` attribute is not set, the test suite's behavior is unchanged. 

Screen Shot 2022-08-11 at 18 32 37
Screen Shot 2022-08-11 at 19 37 30

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -39,6 +39,7 @@ s3transfer = "==0.5.2"
 six = "==1.16.0"
 urllib3 = "==1.26.9"
 typing-extensions = "==4.3.0"
+panther-core = {editable = true, ref = "20d6185cc0d7e4dcf2146528675a7ae034d798dd", git = "https://github.com/panther-labs/panther_core.git"}
 
 [requires]
 python_version = "3.9"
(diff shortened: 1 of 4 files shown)


darwayne:
hmm I think we'll need to make updates in panther-enterprise as well. [Possibly here](https://github.com/panther-labs/panther-enterprise/blob/0ec3e01f30f3eef7466b3945a0712c7527c1aeb1/api/lambda/analysis/models/testing.go#L95)

darwayne:
closing since this has been inactive for a while

PR #209, opened 2022-09-15
Add bulk upload for Config-SDK-based analysis modules

cdzombak:
### Background

This PR implements bulk upload for Config-SDK-based analysis modules. This function is accessible via the `configsdk upload` subcommand, also also runs at the end of the normal `upload` subcommand.

### Changes

* Initial implemention of the `configsdk upload` command
* Run `configsdk upload` automatically after the normal `upload` process if a `panther_content/__main__.py` file is detected

### Caveats

This initial implementation comes with a few caveats:

* We look at `CWD/panther_content/__main__.py` to discover the user's Config-DK-based analysis module. Other paths are not supported.
* The lambda-based API doesn't currently support this feature, so this new function only supports usage via the public API via API token.

### Testing

* [X] Was able to successfully upload our test Config SDK rules repo to staging via `panther_analysis_tool configsdk --api-token XXX --api-host "YYY" upload`, as long as `panther_analysis_tool` is installed in the Config SDK content's virtualenv

### DIFF
--- a/panther_analysis_tool/backend/graphql/sdk_upload.graphql
+++ b/panther_analysis_tool/backend/graphql/sdk_upload.graphql
@@ -5,5 +5,15 @@ mutation SDKUpload($input: UploadDetectionEntitiesInput!) {
             new
             total
         }
+        policies {
+            modified
+            new
+            total
+        }
+        queries {
+            modified
+            new
+            total
+        }
     }
 }
--- a/panther_analysis_tool/backend/graphql/sdk_upload.graphql
+++ b/panther_analysis_tool/backend/graphql/sdk_upload.graphql
@@ -1,0 +1,9 @@
+mutation SDKUpload($input: SDKUploadInput!) {
+    uploadDetectionEntities(input: $input) {
+        queries {
+            modified
+            new
+            total
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


maxrichie5 (approved):
Looks good to me. I know this isn't finished yet but I approve of what is here already 🚀

darwayne (commented):
Looking good so far to me

panther_analysis_tool/backend/graphql/sdk_upload.graphql
```diff
@@ -0,0 +1,9 @@
+mutation SDKUpload($input: UploadDetectionEntitiesInput!) {
+    uploadDetectionEntities(input: $input) {
+        rules {
```
nmartin413, line 3: probably want this to read back stats from queries and polices, too. they weren't originally scoped for this task, but they can be uploaded now, too.

nmartin413 (approved):

panther_analysis_tool/backend/graphql/sdk_upload.graphql
cdzombak: 👍🏻  That should be pretty quick to add.

cdzombak: Done in a2f1e7ed24dfb4df26963e8238c24e25c82f7ddc

maxrichie5 (approved):

darwayne:
not sure why but looks like ci isn't happy

cdzombak:
> not sure why but looks like ci isn't happy

Yeah, I'm not sure either, but I'll dig into it in the morning!

nmartin413 (approved):

maxrichie5:
Fantastic!

PR #211, opened 2022-09-23
Minor Makefile improvements

cdzombak:
### Background

Most importantly, this marks each target as phony, meaning it will still run if/when a file of the same name appears on the filesystem.

I've also added a default `help` target with documentation:

Screen Shot 2022-09-23 at 9 52 27 AM

### Changes

* Mark Makefile targets as phony
* Document each target
* Add default `help` target
* Fix usage of deprecated `pipenv lock -r` option, as in https://github.com/panther-labs/panther_analysis_tool/pull/209/commits/5baf5352942592010d2d74eb9719a25a7c4be77e


darwayne (approved):
🤩 LGTM 🤩

nmartin413 (approved):

PR #213, opened 2022-09-27
[config sdk] adds support for direct lambda invocations.

darwayne:
### Background

Config sdk uploads only work with the public-api this adds support for direct lambda invocations. 

### Changes

* updates lambda client to support config uploads
* updates lambda client to handle json decode better ... just presents the response to the user now .. which will be way more actionable than a cryptic json exception message we were showing before
* adds aws-profile arg to the config upload command
* removes checks around config-sdk that forced api only usage

### Testing

* manual uploads
* unit tests

### DIFF
--- a/panther_analysis_tool/backend/lambda_client.py
+++ b/panther_analysis_tool/backend/lambda_client.py
@@ -16,10 +16,11 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-
+import base64
 import os
 import json
 import logging
+import typing
 
 from typing import Dict, Any, Optional
 from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


cdzombak (approved):

panther_analysis_tool/backend/lambda_client.py
```diff
@@ -42,21 +43,27 @@
...
-
 LAMBDA_CLIENT_NAME = "lambda"
 AWS_PROFILE_ENV_KEY = "AWS_PROFILE"
 
 
+def decode_body(res: BackendResponse) -> typing.Any:
+    try:
+        return json.loads(res.data['body'])
+    except json.decoder.JSONDecodeError as decode_error:
+        raise Exception(res.data['body']) from decode_error
```
maxrichie5, line 54: TIL that raise from exists

maxrichie5 (approved):

PR #214, opened 2022-10-03
maxrichmond: configsdk unit tests

maxrichie5:
### Background

Enables config sdk unit tests to be run via the command `panther_analysis_tool configsdk test` 

### Changes

* Enables config SDK unit tests to be run via `panther_analysis_tool configsdk test`
* `--minimum-tests` supported
* `--skip-disabled-tests` supported
* Will run automatically after classic format tests are ran

### Testing

* These changes were tested with a config test repo with rules, policies, and scheduled rules


danbiwer:
@maxrichie5 You should add

closes https://app.asana.com/0/1202324455056256/1202728292133942

panther_analysis_tool/cmd/configsdk_test.py
```diff
@@ -0,0 +1,280 @@
...
+        detections: The list of detections to be unit tested.
+        min_tests: The minimum number of unit tests each detection needs to have.
+
+    Returns:
+        True if at least one unit test failed, False if all tests pass.
+    """
+    for detection in detections:
+        print(detection.detection_id)
+        if len(detection.filters) == 0:
+            print('    Detection had no filters to test')
```
danbiwer, line 222: nit
```suggestion
            print('\tDetection had no filters to test')
```

danbiwer: unless you're a spaces over tabs person

panther_analysis_tool/cmd/configsdk_test.py
```diff
@@ -0,0 +1,280 @@
...
+import json
+import logging
+from collections import defaultdict
+from functools import reduce
+from typing import Tuple, Final, Dict, List, Any
+
+from panther_analysis_tool.cmd import config_utils
+
+
+class DetectionType(enum.Enum):
```
danbiwer, line 13: *nit

I wonder if we should import these types from panther core

https://github.com/panther-labs/panther_core/blob/main/panther_core/rule.py#L40-L41

panther_analysis_tool/cmd/configsdk_test.py
```diff
@@ -0,0 +1,280 @@
...
+            prg += f'{filt.get_code()} \n\n'
+            prg += f'_filters.append({filt.get_name()}) \n\n'
+
+        # flip the return vals if it is a policy
+        match_val = False
+        if detection_type is DetectionType.POLICY:
+            match_val = True
+        no_match_val = not match_val
+
+        prg += 'def _execute(event):\n'
```
danbiwer, line 55: You're not building an abstract syntax tree from scratch? casual

danbiwer: this is interesting though, so you're basically building out the template defined [here](https://github.com/panther-labs/panther-enterprise/blob/master/cmd/devtools/pantherconfig/configsdk/apiversions/2022-08-17/detection_rule.py.tmpl#L19-L30)


how can we keep these in sync? would a comment suffice?

panther_analysis_tool/cmd/configsdk_test.py
```diff
@@ -0,0 +1,280 @@
...
+    for detection in detections:
+        print(detection.detection_id)
+        if len(detection.filters) == 0:
+            print('    Detection had no filters to test')
+            continue
+
+        for unit_test in detection.unit_tests:
+            prg = unit_test.get_prg(detection.filters, detection.detection_type)
+            locs: Dict[str, str] = {}
+            exec(prg, {}, locs)  # nosec B102 pylint: disable=W0122
```
danbiwer, line 228: I think i'm missing something here, how does the mock event get passed in?  I'm not familiar with config sdk unit tests

panther_analysis_tool/cmd/configsdk_test.py
maxrichie5: I used spaces everywhere instead of tabs because tabs can render different in the terminal and it didn't match up with the way the classic unit test output gets printed to the console

panther_analysis_tool/cmd/configsdk_test.py
maxrichie5: I can. The ones you linked to are string constants though and I was specifically wanting an enum here. So I am not sure it would help at all to import the core ones

panther_analysis_tool/cmd/configsdk_test.py
maxrichie5: Yes exactly. I could add a comment. Idk if it is an issue though for this open source to have a comment relating to our closed source code.

I also don't think it will be an issue since I would imagine once we set these they will never change.

Also the internal template is to support old detection running. Once we switch over completely to the config I think that template will go away

panther_analysis_tool/cmd/configsdk_test.py
maxrichie5: The event is written as a variable as a part of the program. The evaluate function gets called using the variable. The result is saved in the `locs` dictionary

panther_analysis_tool/cmd/configsdk_test.py
danbiwer: good point, maybe we should reference this in panther-enterprise

But yeah, hopefully this never changes.

danbiwer (approved):

panther_analysis_tool/cmd/configsdk_test.py
danbiwer: The idea was to reference what I think is the ultimate source of truth of detection types, but these should never change, so probably not a big deal

PR #223, opened 2022-11-03
Migrate from CircleCI -> GHA

egibs:
### Background

Closes [this](https://app.asana.com/0/1202423273325597/1203272741863901) task.

Moves the CI Workflow from CircleCI to GitHub Actions.

### Changes

* Removes the original CircleCI config 
  * The project will need to be configured to stop building on the CircleCI side to remove the failing check
* Adds a new GitHub Actions Workflow to handle CI
* Adds an `install-pipenv` Makefile target

### Testing

* The new GitHub Action Workflow passes as expected


egibs:
Successful Workflow run can be found [here](https://github.com/panther-labs/panther_analysis_tool/actions/runs/3385411405/jobs/5623522873).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
+on:
+  push:
```
egibs, line 2: `push:` should match the default behavior outlined by CircleCI [here](https://circleci.com/docs/triggers-overview/#run-a-pipeline-on-commit-to-your-code-repository).

> Once a project is hooked up to CircleCI, a pipeline is triggered each time a commit is pushed on a branch that has a `.circleci/config.yml` file included.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
```
tscott8706, line 9: This can be condensed (although you may still prefer the name - several places we use it does not, I don't think), but can we make it an easier version to read? Something like https://github.com/panther-labs/test-automation/blob/main/.github/workflows/ci.yml#L10

Same comment for setting up Python.

Using the v2 or v3 vs what I believe is a commit SHA will be a lot easier to read. Unless you need those specific SHAs for some reason.

.github/workflows/ci.yml
tscott8706: Nit, but not required - I don't think we need the colon after the push

tscott8706 (changes requested):

.github/workflows/ci.yml
egibs: Yea, I can -- I got into the habit of pinning to SHAs since it's one of GitHub's security hardening [recommendations](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) for Actions.

That being said, we aren't really using any obscure Actions or handling sensitive data with these Workflows (and we aren't running on self-hosted runners).

egibs: Addressed in [`f402406` (#223)](https://github.com/panther-labs/panther_analysis_tool/pull/223/commits/f40240698bc222edd382eb215d41f6c2676119f7).

.github/workflows/ci.yml
egibs: Addressed in [`f402406` (#223)](https://github.com/panther-labs/panther_analysis_tool/pull/223/commits/f40240698bc222edd382eb215d41f6c2676119f7).

tscott8706 (approved):

egibs:
> You'll need to "Stop biuld" (I think that's what it's called) in CircleCI and also disable this as a required check (and add the CI GH Actions one as required) to maintain functionality (and commit this). You may not have org. perms... I think I do and can do this. I'll do it whenever you tell me.

You can do that now 👍

tscott8706:
All good - you should be good to merge

cdzombak (approved):

PR #224, opened 2022-11-03
Maxrichmond: rename sdk

maxrichie5:
## References

Relates to https://app.asana.com/0/1202324455056256/1203245566811092/f

## Background

Renaming panther-config/config-sdk to panther-sdk

## Changes

- Version to 0.16.4
- Update `panther_config` to `panther_sdk`
- `config` subcommand becomes `sdk`
- minor fix with `sdk test` result code being backwards

### Testing

* Tested against local clone of `panther_content`


darwayne (dismissed):

panther_analysis_tool/main.py
```diff
@@ -1675,27 +1675,27 @@ def setup_parser() -> argparse.ArgumentParser:
...
     check_conn_parser.set_defaults(func=func_with_backend(check_connection.run))
 
-    # -- config command
+    # -- sdk command
 
-    configsdk_parser = subparsers.add_parser(
-        "config", help="Perform operations using the new Config SDK exclusively "
-                       "(pass config --help for more)"
+    panthersdk_parser = subparsers.add_parser(
+        "sdk", help="Perform operations using the new sdk SDK exclusively "
```
cdzombak, line 1681: `sdk SDK` - looks like a slight find/replace glitch ;)

setup.py
```diff
@@ -1,5 +1,9 @@
+from pathlib import Path
 from setuptools import setup, find_packages
 
+this_directory = Path(__file__).parent
+version = (this_directory / "VERSION").read_text()
```
cdzombak, line 5: I like this, but I would `.strip()` the resulting text. Your new `VERSION` file doesn't contain a trailing newline, but many editors will automatically add one without anyone noticing.

cdzombak (changes requested):
Two very minor changes requested; overall 👍🏻

panther_analysis_tool/main.py
maxrichie5: Thanks!

setup.py
maxrichie5: That's a good call. We will have to add this to our other repos too because I copied this from them.

cdzombak: gpgp. No idea if it will actually cause any issues, but it can't hurt.

PR #228, opened 2022-11-17
Maxrichmond: print nested errors

maxrichie5:
### Background

Closes https://app.asana.com/0/1202324455056256/1202863826116444/f

Print nested errors

### Changes

* Checks for nested error messages 
* Prints errors for nested error messages
* ran `make fmt`

### Testing

* Duplicate rule id with both API and Role auth

#### API

```
...
[INFO]: Uploading items to Panther
[ERROR]: failed to upload to backend: {'message': '{"issues":[{"path":"temp.yml","errorMessage":"duplicate detection id: AWS.WAF.Disassociation"}]}', 'path': ['uploadDetectionEntities'], 'extensions': {'reportable': False, 'statusCode': 400}}
```

#### Role

```
...
[INFO]: Uploading items to Panther
[ERROR]: failed to upload to backend: {'statusCode': 400, 'headers': {}, 'multiValueHeaders': {}, 'body': '{"issues":[{"path":"aws_waf_disassociation.yml","errorMessage":"duplicate detection id: AWS.WAF.Disassociation"}]}'}
```


darwayne:
Noice. Linter is complaining about 2 things. But other than that this is looking good

panther_analysis_tool/backend/client.py
```diff
@@ -156,21 +156,33 @@ def bulk_upload(self, params: BulkUploadParams) -> BackendResponse[BulkUploadRes
...
     @abstractmethod
-    def panthersdk_bulk_upload(self, params: PantherSDKBulkUploadParams) -> BackendResponse[PantherSDKBulkUploadResponse]:
+    def panthersdk_bulk_upload(
+            self, params: PantherSDKBulkUploadParams
+    ) -> BackendResponse[PantherSDKBulkUploadResponse]:
         pass
+
+
+def backend_response_failed(resp: Union[BackendResponse, dict]) -> bool:
+    return resp.status_code >= 400 or resp.data["statusCode"] >= 400  # type: ignore
```
cdzombak, line 188: I'm a little confused about this method. Can you clarify why it accepts `resp: Union[BackendResponse, dict]`? Does the method really work if passed a dict?

maxrichie5: Sorry it can't. I never cleaned this up from when I was developing it

panther_analysis_tool/backend/client.py
```diff
@@ -182,3 +182,7 @@ def panthersdk_bulk_upload(
         self, params: PantherSDKBulkUploadParams
     ) -> BackendResponse[PantherSDKBulkUploadResponse]:
         pass
+
+
+def backend_response_failed(resp: BackendResponse) -> bool:
+    return resp.status_code >= 400 or resp.data["statusCode"] >= 400
```
cdzombak, line 188: Does `resp.data` always contain a `statusCode` key? Is it worth changing to access it here via `.get`, to be cautious?

maxrichie5: I think it does but I will change because better to be cautious

panther_analysis_tool/backend/client.py
```diff
@@ -182,3 +182,7 @@ def panthersdk_bulk_upload(
         self, params: PantherSDKBulkUploadParams
     ) -> BackendResponse[PantherSDKBulkUploadResponse]:
         pass
+
+
+def backend_response_failed(resp: BackendResponse) -> bool:
+    return resp.status_code >= 400 or resp.data.get("statusCode") >= 400
```
cdzombak, line 188:
```suggestion
    return resp.status_code >= 400 or resp.data.get("statusCode", 0) >= 400
```

Otherwise this results in a `TypeError` trying to compare `None` with an `int`

cdzombak (changes requested):

panther_analysis_tool/backend/client.py
maxrichie5: You know I was wondering about that but I figured the linter would have complained so I left it. Thanks for calling it out

cdzombak (dismissed):

cdzombak (approved):

PR #232, opened 2022-12-07
Add SentinelOne schemas

calkim-panther:
### Background

Add sentinelone schemas


maxrichie5 (commented):
Can you update the PAT version here
https://github.com/panther-labs/panther_analysis_tool/blob/master/VERSION

And here
https://github.com/panther-labs/panther_analysis_tool/blob/master/panther_analysis_tool/main.py#L102

maxrichie5 (approved):

PR #240, opened 2023-01-23
fixes bug where our retry logic was not kicking in when another upload was detected.

darwayne:
### Background

Whenever another upload was in progress we weren't properly utilizing our existing retry behavior

### Changes

* Updates lambda client to support upload in progress retry logic
* Updates graphql client to support upload in progress retry logic

### Testing

* unit tests pass
* manually
   - I triggered an two uploads right after each other
   - I did this with both the lambda client and the api client


panther_analysis_tool/backend/lambda_client.py
```diff
@@ -110,7 +110,11 @@ def bulk_upload(self, params: BulkUploadParams) -> BackendResponse[BulkUploadRes
 
         if backend_response_failed(resp):
             err = BackendError(resp.data)
+
             err.permanent = True
+            if "another upload" in resp.data.get("body", ""):
```
maxrichie5, line 115: Can we pluck this logic into a shared function? Whenever we have to do a string check like this in more than one place I think its a good idea to keep that logic in one spot

darwayne: agreed will update

darwayne: updated, let me know if that is what you had in mind

maxrichie5 (approved):

PR #251, opened 2023-02-09
maxrichmond: list bug

maxrichie5:
### Background

I accidentally introduced a bug that caused Dicts to be considered lists

### Changes

* checking isinstance instead of hasattr because dicts are iterable
* added unit test that ensures two version spots are equal

### Testing

* Tested on local repo that was breaking before this change


darwayne (dismissed):

darwayne (commented):
Can we also get a test in so we don't accidentally introduce a regression for this in the future?

maxrichie5:
> Can we also get a test in so we don't accidentally introduce a regression for this in the future?

Good idea

panther_analysis_tool/util.py
```diff
@@ -138,4 +138,4 @@ def deep_get(obj: Dict, path: List[str], default: Any = None) -> Any:
 
 
 def to_list(listish: Any) -> List:
-    return listish if hasattr(listish, "__iter__") else [listish]
+    return listish if isinstance(listish, list) else [listish]
```
cdzombak, line 141: @maxrichie5 does this need to handle other list-ish things (`set`, `tuple`)?

cdzombak: (that'd be `isinstance(listish, tuple)` and `isinstance(listish, set)` IIRC), but of course tests should cover those too.

maxrichie5: I can add those to the tests. It should not. It is only supposed to make a single a list and keep a list a list

cdzombak: Gotcha. I'd like to see a test to document that behavior explicitly, but I'll provisionally approve the PR — just merge once you have that done, please.

cdzombak (dismissed):

panther_analysis_tool/util.py
cdzombak: @maxrichie5 yes! :shipit:

PR #252, opened 2023-02-11
Sorted test results

slw07g:
### Background

Resolves https://github.com/panther-labs/panther_analysis_tool/issues/243

These changes organize the test result output from panther_analysis_tool, such that all relevant details of failed tests are toward the end of the output.

This will allow users to more easily find error messages while debugging detections.

### Changes

* Adds command-line argument `sort-test-results` to panther_analysis_tool
* When set,`sort-test-results` will bucket test results output by passed/errored status, and will sort test cases by Rule ID

### Testing

* Tested locally

PR #257, opened 2023-02-17
Tweak test output when using --sort-test-results; bump to v0.19.7

cdzombak:
### Background

Following up from #252, this PR makes test output when using the `--sort-test-results` option match the test output when not using the option. In particular, this makes any output a test prints appear alongside the test's results.

### Changes

- (92dade0) Print empty line after test result when using --sort-test-results
- (ddacbe7) With --sort-test-results, print test output with the correct result
- (cb799f8) Bump version to v0.19.7

### Testing

* Run Python unit tests
* Run `panther_analysis_tool test` in panther-analysis with and without `--sort-test-results`, comparing the output.


panther_analysis_tool/main.py
```diff
@@ -1239,6 +1264,10 @@ def _run_tests(  # pylint: disable=too-many-arguments
             logging.debug(str(err), exc_info=err)
             failed_tests[detection.detection_id].append(unit_test["Name"])
             continue
+        finally:
```
cdzombak, line 1267: Noting for reviewers that `finally` will run even when an `except` block contains a `continue`: https://docs.python.org/3/reference/compound_stmts.html#finally

maxrichie5 (approved):

PR #262, opened 2023-03-23
add queries summary to output

lindsey-w:
### Background

In investigating another problem, found that we are missing the query summaries in the bulk upload output.  Adding it here.

We are also not printing the final error message when exhausting the retries.  Adding a print here.

### Changes

* print queries summary
* print final error
* update version

### Testing

* api client upload
* lamdba client upload
* `make test`


dekatzenel (approved):
Hard to tell if you've missed a spot, but changes LGTM

dekatzenel (approved):

PR #265, opened 2023-03-31
Update schemas.py

dotbeseck:
Adding in the regex to schemas.py to allow PAT to work

### Background

Testing alerting on Tines.Audit logs resulted in a Regex Error. 

### Changes

* added regex to schemas.py 

### Testing

* run pat test on detections for tines and pass.


darwayne (approved):
Thanks for the contribution @dotbeseck

PR #269, opened 2023-04-18
updates schema to support new optional field

darwayne:
### Background

Adds new optional field to Policies, Rules and Scheduled Rules

### Changes

*  Updates schema to support new field
### Testing

* Unit tests


tests/unit/panther_analysis_tool/test_schemas.py
```diff
@@ -76,3 +78,17 @@ def test_query_rateminutes(self):
             # TimeoutMinutes must be set
             sample_query["Schedule"] = {"RateMinutes": 1}
             SCHEDULED_QUERY_SCHEMA.validate(sample_query)
+
+    def test_rba_flag(self):
```
dekatzenel, line 82: Also validate with absent?

darwayne: yup good call out will do

maxrichie5 (approved):

PR #273, opened 2023-04-20
Split ci and fmt into separate workflows with different targets

dekatzenel:
### Background

We want fmt to operate on pull_request_target to allow fmt changes to be committed to forked PRs, but we do not want ci to operate on pull_request_target because this requires that tests pass on the tip of master which is not guaranteed because the tests depend on changing resources.

### Changes

* Move fmt job from ci.yml to fmt.yml
* Add pull_request_target to fmt.yml

### Testing

* Testing will be GitHub CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,23 +23,3 @@ jobs:
       - name: Run CLI Tests
         run: make ci
 
-  fmt:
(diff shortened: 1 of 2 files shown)


egibs (approved):

darwayne (approved):
🥇 LGTM 🥇

PR #274, opened 2023-04-20
Dana katzenelson lowercase compare per customer request

dekatzenel:
### Background

One last improvement - case insensitive table name comparison
closes: https://app.asana.com/0/1202324455056256/1204427814029085/f

### Changes

* Case insensitive table name comparison

### Testing

* Passes existing tests


darwayne (approved):
Can we also get a new test case going that captures this scenario

PR #275, opened 2023-04-21
Remove rule suppressions schema support and fix tests accordingly

dekatzenel:
### Background

PAT schema validation allows for Suppressions on Rules, but these are unused and unsupported by Panther

### Changes

* Remove support for Suppressions from the Rule Schema
* Modify Tests accordingly

### Testing

* Ensure existing tests pass


darwayne (approved):

dekatzenel:
I didn't bump the version because I figured this could just go out as part of the next change, please let me know if you disagree

maxrichie5 (approved):

dekatzenel:
> > I didn't bump the version because I figured this could just go out as part of the next change, please let me know if you disagree
> 
> If we have pending changes coming yeah it can wait ... if we don't .. might make sense to bump so we don't forget

The TableName improvements are in Slated for Next Sprint, so I think we have pending changes

PR #276, opened 2023-04-25
Add --valid-table-names option for a user-supplied list of valid table names

dekatzenel:
### Background
Customers that manage their own snowflake want the option to specify valid tables that aren't panther-managed in saved queries
closes: https://app.asana.com/0/1202324455056256/1204437125463858/f

### Changes

* Add --valid-table-names option which takes a list of valid table names with wildcard expansion

### Testing

* Added unit and integration testing of the new flag
* Manually tested new flag with .panther_settings.yml file
* Ensured all existing tests pass


panther_analysis_tool/constants.py
```diff
@@ -16,7 +16,7 @@
 
 PACKAGE_NAME: Final = "panther_analysis_tool"
 
-VERSION_STRING: Final = "0.20.1"
+VERSION_STRING: Final = "0.20.2"
```
darwayne, line 19: since this is a new behavior. This should probably go under 0.21.0

darwayne (approved):

panther_analysis_tool/constants.py
dekatzenel: I considered that and wound up (incorrectly) leaning towards "it's tightly coupled to an existing feature"

darwayne: got it, yeah we haven't been super strict about this, but since we're adding a new command argument as well is what pushed me towards version bump. Ideally minor bumps should be just for bug fixes.

darwayne: yup patch, lol bad choice of words there 🤦🏾‍♂️

darwayne (approved):

maxrichie5 (approved):
Awesome work!

Everything looks fine to me. Nice tests!

dekatzenel:
Confirmed with customer that this should meet their needs

PR #281, opened 2023-04-28
adds async bulk upload support for api users

darwayne:
### Background

When using the api token upload requests are limited to 30seconds. Since they are cases where an upload can take more than 30 seconds we need a way to allow users to continue to upload without any timeout issues.

### Changes
* Updates upload requests that are using the api token to use async uploads
* Adds new --no-async flag to the upload command
* Add logic to detect if panther instance supports async bulk uploads
* Fixes bug where we would always retry uploads vs only retrying when needed

### Testing

* Lots of manual tests

### DIFF
--- a/panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
+++ b/panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
@@ -1,0 +1,32 @@
+query status($input: ID!) {
+    detectionEntitiesUploadStatus(receiptId: $input) {
+        status
+        error
+        result {
+            dataModels {
+                ...UploadStatisticDetails
+                }
+                globalHelpers {
+                ...UploadStatisticDetails
+                }
+                lookupTables {
+                ...UploadStatisticDetails
+                }
+                policies {
+                ...UploadStatisticDetails
+                }
+                rules {
+                ...UploadStatisticDetails
+                }
+                queries {
+                ...UploadStatisticDetails
+                }
+        }
+    }
+}
+
+fragment UploadStatisticDetails on UploadStatistics {
+  modified
+  new
+  total
+}
--- a/panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
+++ b/panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
@@ -5,22 +5,22 @@ query status($input: ID!) {
         result {
             dataModels {
                 ...UploadStatisticDetails
-                }
-                globalHelpers {
+            }
+            globalHelpers {
                 ...UploadStatisticDetails
-                }
-                lookupTables {
+            }
+            lookupTables {
                 ...UploadStatisticDetails
-                }
-                policies {
+            }
+            policies {
                 ...UploadStatisticDetails
-                }
-                rules {
+            }
+            rules {
                 ...UploadStatisticDetails
-                }
-                queries {
+            }
+            queries {
                 ...UploadStatisticDetails
-                }
+            }
         }
     }
 }
--- a/panther_analysis_tool/backend/graphql/introspection_query.graphql
+++ b/panther_analysis_tool/backend/graphql/introspection_query.graphql
@@ -1,0 +1,88 @@
+query IntrospectionQuery {
+    __schema {
+      queryType { name }
+      mutationType { name }
+      subscriptionType { name }
+      types {
+        ...FullType
+      }
+      directives {
+        name
+        description
+        locations
+        args {
+          ...InputValue
+        }
+      }
+    }
+  }
+  fragment FullType on __Type {
+    kind
+    name
+    description
+    fields(includeDeprecated: true) {
+      name
+      description
+      args {
+        ...InputValue
+      }
+      type {
+        ...TypeRef
+      }
+      isDeprecated
+      deprecationReason
+    }
+    inputFields {
+      ...InputValue
+    }
+    interfaces {
+      ...TypeRef
+    }
+    enumValues(includeDeprecated: true) {
+      name
+      description
+      isDeprecated
+      deprecationReason
+    }
+    possibleTypes {
+      ...TypeRef
+    }
+  }
+  fragment InputValue on __InputValue {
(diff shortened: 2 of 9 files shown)


panther_analysis_tool/main.py
```diff
@@ -305,7 +305,11 @@ def upload_analysis(backend: BackendClient, args: argparse.Namespace) -> Tuple[i
         A tuple of return code and the archive filename.
     """
 
-    if args.batch:
+    supports_async = backend.supports_async_uploads()
+    if supports_async and args.no_async:
+        supports_async = False
```
dekatzenel, line 310: Would something like the following be better:
`use_async = (not args.no_async) and backend.supports_async_uploads()`

and then use `use_async` below?

panther_analysis_tool/backend/public_api_client.py
```diff
@@ -347,6 +378,28 @@ def _execute(
...
+        request: DocumentNode,
+        variable_values: Optional[Dict[str, Any]] = None,
+    ) -> ExecutionResult:
+        try:
+            res = self._execute(request, variable_values=variable_values)
+        except TransportQueryError as e:  # pylint: disable=C0103
+            err = PermanentBackendError(e)
+            if e.errors and len(e.errors) > 0:
+                err = PermanentBackendError(e.errors[0])
+                err.permanent = not is_retryable_error(e.errors[0])
```
dekatzenel, line 392: Does this mean that you can have a PermanentBackendError with permanent == False? That doesn't make sense to me

panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
```diff
@@ -0,0 +1,32 @@
...
+                policies {
+                ...UploadStatisticDetails
+                }
+                rules {
+                ...UploadStatisticDetails
+                }
+                queries {
+                ...UploadStatisticDetails
+                }
+        }
```
maxrichie5, line 24: Are we able to format this better?

panther_analysis_tool/backend/graphql/introspection_query.graphql
maxrichie5, line 1: What is this file for?

panther_analysis_tool/backend/public_api_client.py
```diff
@@ -127,40 +140,40 @@ def check(self) -> BackendCheckResponse:
...
+            if status == "FAILED":
+                if is_retryable_error_str(error):
+                    raise BackendError(error)
+                raise PermanentBackendError(error)
+            if status == "":
+                raise BackendError("empty data")
+
+            if status == "COMPLETED":
+                return to_bulk_upload_response(data)
+            time.sleep(5)
```
maxrichie5, line 168: Can we move this to the top of this loop so it sleeps before the first check?

Can we also change this to `2` maybe? `5` seems too long

panther_analysis_tool/backend/public_api_client.py
```diff
@@ -127,40 +140,40 @@ def check(self) -> BackendCheckResponse:
...
+            result = res.data.get("detectionEntitiesUploadStatus", {})  # type: ignore
+            status = result.get("status", "")
+            error = result.get("error")
+            data = result.get("result")
+            if status == "FAILED":
+                if is_retryable_error_str(error):
+                    raise BackendError(error)
+                raise PermanentBackendError(error)
+            if status == "":
+                raise BackendError("empty data")
```
maxrichie5, line 164: I feel like this could be a more descriptive error. Maybe like `no bulk upload status available`?

panther_analysis_tool/main.py
```diff
@@ -347,7 +353,10 @@ def upload_zip(backend: BackendClient, args: argparse.Namespace, archive: str) -
 
         while True:
             try:
-                response = backend.bulk_upload(upload_params)
+                if use_async:
+                    response = backend.async_bulk_upload(upload_params)
+                else:
+                    response = backend.bulk_upload(upload_params)
```
maxrichie5, line 359: I might have missed this somewhere in the PR, but do we have logic that says we should only do async upload if we are using the public api?

maxrichie5 (commented):
Couple questions

panther_analysis_tool/main.py
darwayne: ooh I like that. Will update

panther_analysis_tool/backend/public_api_client.py
darwayne: good point, I'll update this one to a BackendError. I added the new `PermanentBackendError` type because the `BackendError` type by default has permanent set to False .. there were some bugs introduced due to that behavior so wanted to be explicit

panther_analysis_tool/backend/graphql/introspection_query.graphql
darwayne: I'm using this to check if the panther instance supports the new async bulk upload endpoint in graphql. If it doesn't we fall back to the old behavior. This should allow us to make a PAT release without waiting for all panther instances to get the latest version

panther_analysis_tool/backend/public_api_client.py
darwayne: sure thing

panther_analysis_tool/main.py
darwayne: yup that case is handled. I just added a supports_async_bulk_upload method on the various backend clients .. the lambda client will default to no. and the graphql client will be true if the instance supports it.

panther_analysis_tool/backend/public_api_client.py
darwayne: good point will update

panther_analysis_tool/backend/public_api_client.py
darwayne: done

panther_analysis_tool/backend/public_api_client.py
darwayne: done

dekatzenel (approved):

PR #283, opened 2023-05-04
Removing inaccurate note about config file

eliskaachee:
### Background

The config file does not override command line values. So removing text that says it does.

### Changes

Removing text.

### Testing

Is text removed? Then yay!


example_panther_config.yml
```diff
@@ -1,6 +1,5 @@
 # Example configuration file for panther_analysis_tool.
 # Copy this file into the root of your test repository and name it .panther_settings.yml
-# Note that options in this file override options passed on the command line
```
darwayne, line 3: Can we just update the note to be accurate?

panther_analysis_tool/main.py
```diff
@@ -1775,11 +1775,7 @@ def run() -> None:
         if key.startswith("PANTHER_"):
             logging.info("Found Environment Variables prefixed with 'PANTHER'.")
             break
-    if os.path.exists(CONFIG_FILE):
-        logging.info(
-            "Found Config File %s . NOTE: SETTINGS IN CONFIG FILE OVERRIDE COMMAND LINE OPTIONS",
```
darwayne, line 1780: same here, can we keep this but just say what it is actually doing instead?

dekatzenel: Maybe check if any command line options are provided, if it's not too difficult, and only mention the override if so

example_panther_config.yml
eliskaachee: Okay, how about something like "Note that options in this file will be overwritten by options passed on the command line"?

panther_analysis_tool/main.py
eliskaachee: That's a good idea, how about something like "Found Config File ____ NOTE: COMMAND LINE OPTIONS WILL OVERWRITE SETTINGS IN CONFIG FILE"

For checking if any command line options are provided, is that just checking "if args" from line 1758? Or do we just want to check for specific args?

example_panther_config.yml
dekatzenel: Nit: s/overwritten/overridden but otherwise LGTM

panther_analysis_tool/main.py
dekatzenel: > how about something like "Found Config File ____ NOTE: COMMAND LINE OPTIONS WILL OVERWRITE SETTINGS IN CONFIG FILE"
same overwrite vs override switch but LGTM
> For checking if any command line options are provided, is that just checking "if args" from line 1758? Or do we just want to check for specific args?
Looks like args includes the subcommand which isn't in the config file, so you'd need to check if it has more args than that

dekatzenel (approved):

PR #284, opened 2023-05-11
jsonschema validation of simple detections

lindsey-w:
### Background

Adding jsonschema validation for simple detections to panther_analysis_tool.  This _keeps_ the existing schema validation on top of the jsonschema validation, until we can migrate the remaining validations into the jsonschema itself.

### Changes

* jsonschema added
* validate against the schema in the case of simple detection or filters 
* tests

### Testing

* `make test`
* pip install from source + validate


panther_analysis_tool/util.py
```diff
@@ -173,3 +173,9 @@ def convert_unicode(obj: Any) -> str:
     e.g. \\\\u003c => <"""
     string_to_convert = str(obj)
     return re.sub(r"\\*\\u([0-9a-f]{4})", lambda m: chr(int(m.group(1), 16)), string_to_convert)
+
+
+def is_simple_detection(analysis_item: Dict[str, Any]) -> bool:
+    return (
+        analysis_item.get("Detection") is not None or analysis_item.get("InlineFilter") is not None
```
darwayne, line 180: hmm should we leave out the InlineFilter part until we have support for that? IIRC presence of inline filters does not guarantee that the detection is a simple detection

lindsey-w: Done [77084ba](https://github.com/panther-labs/panther_analysis_tool/pull/284/commits/77084ba3d16fcc3d2168939116ebe7951e536868)

darwayne: thanks!

darwayne (approved):

maxrichie5 (approved):
LGTM

In the long run, I think it will be better to post the schema to the public bucket or something, then consume it from there. That way customer's can use it in their IDE too

PR #297, opened 2023-06-02
Add Notion to schema

grantjoy:


andrea-youwakim (approved):
adding review from security side, please wait for an approval from detections as well prior to merging if possible

maxrichie5 (approved):

PR #298, opened 2023-06-05
refactoring: classified analysis container

maxrichie5:
### Background

We have been using a map but treating it as a class. We have been using a collection of constant strings to interact with that map. This has lead to more error prone code that is harder to read.

### Changes

* Creates `ClassifiedAnalysisContainer` class to replace map
* Namespaces all analysis type constants under `AnalysisTypes` class

### Testing

* All existing tests pass
* panther-analysis tests all still pass
* Uploading panther-analysis works
* Uploading panther-analysis in batches works

### DIFF
--- a/panther_analysis_tool/analysis_utils.py
+++ b/panther_analysis_tool/analysis_utils.py
@@ -79,17 +79,17 @@ class ClassifiedAnalysisContainer:
         return all(len(l) == 0 for l in self._self_as_list())
 
     def apply(
-        self,
-        fn: Callable[[List[ClassifiedAnalysis]], List[ClassifiedAnalysis]],
+            self,
+            func: Callable[[List[ClassifiedAnalysis]], List[ClassifiedAnalysis]],
     ) -> "ClassifiedAnalysisContainer":
         container = ClassifiedAnalysisContainer()
-        container.data_models = fn(self.data_models)
-        container.globals = fn(self.globals)
-        container.detections = fn(self.detections)
-        container.simple_detections = fn(self.simple_detections)
-        container.scheduled_queries = fn(self.scheduled_queries)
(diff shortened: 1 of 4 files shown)


panther_analysis_tool/analysis_utils.py
```diff
@@ -47,6 +48,56 @@ def __init__(self, file_name: str, dir_name: str, analysis_spec: Dict[str, Any])
         self.analysis_spec = analysis_spec
 
 
+@dataclasses.dataclass
+class ClassifiedAnalysisContainer:
```
maxrichie5, line 52: This is the best name I could think of. I am open to suggestions if you can think of a better one!

panther_analysis_tool/main.py
```diff
@@ -1055,21 +1044,32 @@ def classify_analysis(
...
+            elif analysis_type == AnalysisTypes.DATA_MODEL:
+                all_specs.data_models.append(classified_analysis)
+            elif analysis_type == AnalysisTypes.GLOBAL:
+                all_specs.globals.append(classified_analysis)
+            elif analysis_type == AnalysisTypes.LOOKUP_TABLE:
+                all_specs.lookup_tables.append(classified_analysis)
+            elif analysis_type == AnalysisTypes.PACK:
+                all_specs.packs.append(classified_analysis)
+            elif analysis_type == AnalysisTypes.SCHEDULED_QUERY:
+                all_specs.scheduled_queries.append(classified_analysis)
```
darwayne, line 1072: `[minor]` hmm this logic feels like it might be better suited being in that new container class you created

maxrichie5: Good point. I will move it

dekatzenel (approved):

PR #299, opened 2023-06-06
Add character limit check for schema names

papanikge:
### Background

Schema names cannot be longer than 255 characters.
We already handle this case in the BE, but we can avoid sending the request whatsoever.

### Changes

* Adds character limit check for schema names

### Testing

* `panther_analysis_tool update-custom-schemas --path <custom_path>`

### Screenshots 

image


panther_analysis_tool/log_schemas/user_defined.py
```diff
@@ -203,6 +203,9 @@ def _extract_schema_name(
                 f" with the prefix '{self._SCHEMA_NAME_PREFIX}'",
             )
 
+        if len(name) > 255:
+            return ("", f"'schema' field: value should not exceed 255 characters")
```
darwayne, line 207: linter doesn't like that you're using an f string without variable here

```suggestion
            return ("", "'schema' field: value should not exceed 255 characters")
```

darwayne:
LGTM once CI is happy

darwayne (approved):

PR #301, opened 2023-06-07
testing of inline filters

maxrichie5:
### Background

Users want to be able to test InlineFilters along with their detections to match the testing flow of the console and backend. This will test inline filters that appear on rules for both simple detections and python detections.

### Changes

* Calls backend to transpile filters when they exist on detections
* Adds filters to the running of unit tests for detections

### Testing

* More unit tests
* Manually tested on panther analysis


panther_analysis_tool/main.py
```diff
@@ -903,39 +907,27 @@ def setup_run_tests(  # pylint: disable=too-many-locals,too-many-arguments
         if skip_disabled_tests and not analysis_spec.get("Enabled", False):
             continue
         analysis_type = analysis_spec["AnalysisType"]
-        analysis_id = analysis_spec.get("PolicyID") or analysis_spec["RuleID"]
+
+        detection_args = dict(
+            id=analysis_spec.get("PolicyID") or analysis_spec["RuleID"],
+            analysisType=analysis_type.upper(),
```
maxrichie5, line 913: The analysis type needs to be upper case to work in panther core. I am not sure how this was working before without that

darwayne (commented):
nice this looks good for the most part .. the new transpile endpoint will be in next week's release right? If a user pulls down PAT prior to that release will they just see a warning?

dekatzenel (approved):

maxrichie5:
> Yes it will be in next week's release. Since we now release in tangent it should be all good. If a user is using filters prior to that they will see a warning. If the user doesn't have any filters it will skip it altogether with no warning

Oops. Looks like I put a bug in that logic. Thanks for the callout Dar! Pushing fix...

darwayne:
hmm looks like CI still isn't happy?

maxrichie5:
> hmm looks like CI still isn't happy?

I know. I am waiting for a fix to make its way to github. It is just reverting test changes I didn't need though so still good to review

darwayne (approved):

PR #305, opened 2023-06-21
adds support for validate command.

darwayne:
### Background
This adds a new validate command to PAT. Note this change depends on https://github.com/panther-labs/panther-enterprise/pull/13425

**Validation success**
image

**Validation Failure**
image

**New API endpoint not detected**
image

### Changes

* Adds new validate command
* Updates client to not validate requests against graphql schema
    * protected apis are not visible in the schema
### Testing

* Manual validation locally
* Unit tests


panther_analysis_tool/backend/public_api_client.py
```diff
@@ -185,6 +193,32 @@ def bulk_upload(self, params: BulkUploadParams) -> BackendResponse[BulkUploadRes
 
         return to_bulk_upload_response(data)
 
+    def bulk_validate(self, params: BulkUploadParams) -> BulkUploadValidateStatusResponse:
```
dekatzenel, line 196: I realize that async bulk upload is the same, but will this just go forever if the status never changes (e.g. there's an error in a way that prevents the status from changing)?

dekatzenel (commented):
Generally LGTM, but we need to wait until rest of work is done to merge

panther_analysis_tool/backend/public_api_client.py
darwayne: good call out I'll get a fix in for that

darwayne: fixed

dekatzenel: This is a good fix, but not what I was thinking of. I meant some sort of cap on retries, in case the status got stuck forever as NOT_PROCESSED

darwayne: ooh, that's enforced on the backend api. currently set at 20minutes. If we want to change that behavior I think that falls out of the scope of this PR

panther_analysis_tool/backend/client.py
```diff
@@ -33,6 +33,10 @@ class PermanentBackendError(BackendError):
     permanent: bool = True
 
 
+class UnSupportedEndpointError(Exception):
```
maxrichie5, line 36:
```suggestion
class UnsupportedEndpointError(Exception):
```

maxrichie5 (dismissed):

panther_analysis_tool/backend/client.py
darwayne: nice catch let me fix that up

dekatzenel:
@darwayne didn't we want this to wait for metrics?

darwayne:
@dekatzenel yeah we can

dekatzenel (approved):

darwayne:
We decided to not let this be blocked by the analytics work

PR #314, opened 2023-07-10
Use panther-bot-automation PAT for pushing fmt changes

mDemianchuk:
### Background

This PR replaces the external GHA [stefanzweifel/git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs) and addresses the repository's `GITHUB_TOKEN` limitation:
> When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.

Because of this, if the final commit was made from GHA, the PR will be stuck waiting for workflows that will never trigger.

Closes https://app.asana.com/0/1202325082509077/1205000151188983/f

### Note

This PR only addresses the described problem in branches / PRs that are made in this repo. 

If forks, the GHA will fall back to the repository's token, and because of this, it will not trigger workflows on the automated commit -- the final commit must be made by the user


.github/workflows/fmt.yml
```diff
@@ -1,11 +1,24 @@
-on: [push, pull_request, pull_request_target]
+on: push
```
mDemianchuk, line 1: Given the nature of this workflow, I think it's okay to run it on `push` only -- this should also cover pushes made in fork PRs

.github/workflows/fmt.yml
```diff
@@ -1,11 +1,24 @@
...
+      - name: Configure git user
+        run: |
+          if [[ -n "${{ secrets.PANTHER_BOT_AUTOMATION_TOKEN }}" ]]; then
+            echo "GIT_USER_NAME=panther-bot-automation" >> $GITHUB_ENV
+            echo "GIT_USER_EMAIL=github-service-account-automation@panther.io" >> $GITHUB_ENV
+            echo "GIT_TOKEN=${{ secrets.PANTHER_BOT_AUTOMATION_TOKEN }}" >> $GITHUB_ENV
+          else
+            echo "GIT_USER_NAME=${{ github.actor }}" >> $GITHUB_ENV
+            echo "GIT_USER_EMAIL=$(git log -n 1 --pretty=format:%ae)" >> $GITHUB_ENV
+            echo "GIT_TOKEN=${{ github.token }}" >> $GITHUB_ENV
```
mDemianchuk, line 16: We will fall back to the default token in fork PRs, and because of this the workflow will not be re-triggered

.github/workflows/fmt.yml
egibs: Do you want to exclude the default branch if we're running on every `push` event?

mDemianchuk: Good call! I'll add that

PR #319, opened 2023-07-12
Allow additionalProperties on top level schema and cut panther_analysis_tool 0.24.2

zacbrown:
### Background

PAT 0.24 introduced additional schema validation in the form of JSON schema validation. However, this conflicts with our flag `--ignore-extra-keys`. This change updates the jsonschema and updates test content for this change.

### Changes

- updated the schema to be permissive of additional top level keys
- added a test case demonstrating the additional top level key
- modified the behavior of some test cases to force `--ignore-extra-keys` given the permissiveness of the schema now
- bumped version number to 0.24.2

### Testing

* unit tests
* manual testing of PAT


edyesed (approved):
formatting changes on analysis jsonschema and ignoring extra keys when testing

darwayne (approved):

PR #321, opened 2023-07-17
Improve JSON validation performance

corrylc:
Existing code (likely introduced in #311) validated the JSON schema, `ANALYSIS_CONFIG_SCHEMA`, on every loop. This was excessive since once is clearly enough for an unchanging specification (and even once is excessive since the upstream schema should be presumed accurate at unit test time).

Therefore we create the JSON validator outside the loop, then re-use it on every run. Use of `Draft202012Validator` is based on a _quick_ reading of `jsonschema` docs. There is a newer draft that may be more appropriate, but this validator works as well.

Overall, this reduces runtime on a MacBook Pro M2 from 45 seconds to 5 seconds (about 9x improvement). Broadly this restores the tool to its typical runtime.


darwayne:
Thanks for the contribution @corrylc. Let me update your branch and see if it will make CI happy

nhakmiller (approved):

darwayne:
@corrylc got your changed merged in via a separate PR. Our team will investigate the CI issues you ran into.

PR #328, opened 2023-07-25
Improve JSON validation performance

darwayne:
Existing code validated the JSON schema, `ANALYSIS_CONFIG_SCHEMA`, on every loop. This was excessive since once is clearly enough for an unchanging specification.

Therefore we create the JSON validator outside the loop, then re-use it on every run.

Overall, this reduces runtime on a MacBook Pro M2 from 45 seconds to 5 seconds (about 9x improvement).

**Note**
This change was initially proposed in #321 .. creating a separate PR as a short term work around for CI issues


nhakmiller (approved):

panther_analysis_tool/main.py
```diff
@@ -1001,6 +1002,9 @@ def classify_analysis(
     # add any duplicates to the invalid_specs
     analysis_ids: List[Any] = []
 
+    # Create a json validator and check the schema only once rather than during every loop
+    json_validator = Draft202012Validator(ANALYSIS_CONFIG_SCHEMA)
```
dekatzenel, line 1006: Why hardcoded rather than using the `jsonschema.validator_for()` method?

dekatzenel (approved):
One question

PR #335, opened 2023-08-03
Implement benchmark subcommand

dekatzenel:
### Background
Add `benchmark` subcommand to support performance testing in the style of Data Replay
closes: https://app.asana.com/0/1202324455056256/1205095851037499/f

### Changes

* Add `benchmark` subcommand which validates inputs and then repeatedly performance tests a rule against a log_type for a specific hour of historical data
* Rename `cmd` dir to `command` to match test version

### Testing

* Added unit tests
* Manually tested command against dev instance


panther_analysis_tool/backend/public_api_client.py
```diff
@@ -457,6 +475,81 @@ def supports_async_uploads(self) -> bool:
...
+                "hour": params.hour.astimezone().isoformat(),
+                "logType": params.log_type,
+            }
+        }
+        res = self._potentially_supported_execute(query, variable_values=create_params)
+        replay_id = res.data.get("createPerfTest", {}).get("replay", {}).get("id")  # type: ignore
+        if not replay_id:
+            raise BackendError("empty data")
+        stopped = False
+        terminal_statuses = ["DONE", "CANCELED", "ERROR_EVALUATION", "ERROR_COMPUTATION"]
```
darwayne, line 529: we should probably make these constants since we use them in multiple spots

panther_analysis_tool/command/benchmark.py
```diff
@@ -0,0 +1,281 @@
...
+    for _ in range(0, args.iterations):
+        replay_response = backend.run_perf_test(
+            PerfTestParams(
+                zip_bytes=buffer.read(),
+                log_type=log_type,
+                hour=hour_or_err,
+                timeout=timeout.astimezone(),
+            )
+        )
+        if replay_response.data.state == "CANCELED":
```
darwayne, line 79: we should probably make all the various states constants

panther_analysis_tool/command/benchmark.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+def validate_hour(
+    args: argparse.Namespace, log_type: str, backend: BackendClient
+) -> Union[datetime.datetime, str]:
+    hour = getattr(args, "hour", None)
+    now_truncated = datetime.datetime.now().replace(
+        minute=0, second=0, microsecond=0
+    ).astimezone() + datetime.timedelta(microseconds=-1)
+    window_begin = now_truncated + datetime.timedelta(weeks=-2, microseconds=1)
```
darwayne, line 147: hmm I wonder if we should try and get this from the backend since this number can change and has changed over the last few months.

darwayne (approved):
Looking good 🤩 

Had some minor non blocking comments

panther_analysis_tool/command/benchmark.py
dekatzenel: Makes sense - do you know if it's currently retrievable?

darwayne: I think the UI gets it via an api call, but yeah to get it might require a new protected api endpoint

panther_analysis_tool/constants.py
```diff
@@ -87,3 +87,12 @@ class AnalysisTypes:
 ]
 
 BACKEND_FILTERS_ANALYSIS_SPEC_KEY = "_backend_filters"
+
+
+class ReplayStatus:
```
darwayne, line 92: 👍🏾

panther_analysis_tool/command/benchmark.py
dekatzenel: Recorded under M2: https://app.asana.com/0/1205095847919624/1205187529819686/f

darwayne: sweet thanks

PR #340, opened 2023-08-08
adds support for additional log types.

darwayne:
### Background
We noticed that we were missing the AWS.Config log type. So went ahead and double checked we weren't missing any other  log types.

### Changes
* Updates log types to match with what panther currently offers

### Testing

* Got a list of all log types panther currently supports and added it to a unit test


dekatzenel (approved):
Looking forward to when we validate this against the BE instead

PR #339, opened 2023-08-07
Display subcommand help text on subcommand help (as description)

dekatzenel:
### Background

Ensure that the text that displays per-subcommand when running `panther_analysis_tool --help` also displays when running `panther_analysis_tool subcommand --help`. Provides useful context about the subcommands

### Changes

* duplicate subparser help as description
* add arg default values to help output

### Testing

* Manually checked that output appeared

PR #326, opened 2023-07-25
Allowing users to manage field discovery schema property through PAT

kostaspap:
### Background

<High level overview here>

### Changes

* <Describe changes here>

### Testing

* <Testing steps>

PR #341, opened 2023-08-08
Release 0.25.0

maxrichie5:
### Background

Releasing 0.25.0 

### Changes

* New Tenable Vulnerability Log Type (#334)
* Managed Field discovery (#326)
* Benchmark command (#335...#339)
* Fix various log types missing from validation (#340)


edyesed:
Lines 79-82 and line 99 from #340 seem to be for lookup tables, which I think aren't strictly log types. 

Probably no harm in having them included, and I do not know whatever consequences may exist if somebody were to create a detection that had this as the input LogType.

edyesed (approved):

darwayne:
hmm good call out @edyesed it is an option in the UI today. If that's wrong we should probably fix that
image

edyesed:
> hmm good call out @edyesed it is an option in the UI today for log types. If that's wrong we should probably fix that image

I did not realize they surfaced there as well. Maybe we should check in with the Data team. If they're currently in the UI, then that is great because this brings panther_analysis_tool to parity 💪

darwayne:
Yeah I think we're good to merge this as is. If we find out these are no bueno we can fix them in a follow up.

darwayne (approved):

PR #344, opened 2023-08-09
Adding Proofpoint.Event

grantjoy:
New schema.


maxrichie5:
Hi Grant. Can we wait on this? I just made a new tool for making this easier and this would be a great first test https://github.com/panther-labs/panther-enterprise/pull/14454

grantjoy:
@maxrichie5 sure, no problem. this doesn't actually go live until Aug 22nd, I will let you handle the change.

PR #346, opened 2023-08-10
Introduce new enrich-test-data command

zacbrown:
## Background
This change introduces a new `panther_analysis_tool` command `enrich-test-data`. This functionality mirrors the Panther Console (web app) experience for enriching test data in Detection unit tests.

### Changes

* New command to query Panther for enriched test data based on an input test case.
* If enrichment was added to a test case, the Analysis Spec YAML is updated. These changes can be noisy in some cases:
  * `Log` or `Resource` content in `Tests` are inline JSON instead of YAML.
  * YAML is not formatted according to standard YAML formatting (2 spaces for indentation).
  * Superfluous newlines that do not match typical YAML formatting - e.g. newline between list item dash `-` and a sequence mapping declaration.

### Testing

* Tests for core functionality of `enrich-test-data`
* Manual testing on the entirety panther-analysis.

### DIFF
--- a/tests/unit/panther_analysis_tool/test_enriched_event_generator.py
+++ b/tests/unit/panther_analysis_tool/test_enriched_event_generator.py
@@ -18,7 +18,7 @@ def get_specs_for_test() -> typing.Dict[str, LoadAnalysisSpecsResult]:
         AnalysisTypes.RULE: LoadAnalysisSpecsResult(
             f"filname.rule",
             f"filepath.rule",
-            get_yaml_loader().load(
+            get_yaml_loader(roundtrip=True).load(
                 """
                 RuleID: foo.bar.rule
                 AnalysisType: rule
@@ -32,13 +32,13 @@ def get_specs_for_test() -> typing.Dict[str, LoadAnalysisSpecsResult]:
                         json: {"foo": "bar"}
                 """
             ),
-            yaml_ctx=get_yaml_loader(),
+            yaml_ctx=get_yaml_loader(roundtrip=True),
             error=None
         ),
         AnalysisTypes.SCHEDULED_RULE: LoadAnalysisSpecsResult(
             f"filname.scheduled_rule",
             f"filepath.scheduled_rule",
-            get_yaml_loader().load(
+            get_yaml_loader(roundtrip=True).load(
                 """
                 RuleID: foo.bar.scheduled_rule
                 AnalysisType: scheduled_rule
@@ -52,13 +52,13 @@ def get_specs_for_test() -> typing.Dict[str, LoadAnalysisSpecsResult]:
                         json: {"foo": "bar"}
                 """
             ),
-            yaml_ctx=get_yaml_loader(),
+            yaml_ctx=get_yaml_loader(roundtrip=True),
             error=None
         ),
         AnalysisTypes.POLICY: LoadAnalysisSpecsResult(
             f"filname.policy",
             f"filepath.policy",
-            get_yaml_loader().load(
+            get_yaml_loader(roundtrip=True).load(
                 """
                 PolicyID: foo.bar.policy
                 AnalysisType: policy
@@ -72,7 +72,7 @@ def get_specs_for_test() -> typing.Dict[str, LoadAnalysisSpecsResult]:
                         json: {"foo": "bar"}
                 """
             ),
-            yaml_ctx=get_yaml_loader(),
+            yaml_ctx=get_yaml_loader(roundtrip=True),
             error=None
         )
     }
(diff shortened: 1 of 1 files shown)


panther_analysis_tool/enriched_event_generator.py
```diff
@@ -0,0 +1,180 @@
...
+                    new_list.append(
+                        EnrichedEventGenerator._convert_inline_json_dict_to_python_dict(item)
+                    )
+                else:
+                    new_list.append(item)
+            return new_list
+        return data
+
+    def _handle_analysis_item(self, analysis_id: str, test: dict, test_case_field_key: str) -> dict:
+        if test_case_field_key not in test:
```
darwayne, line 93: we should also check to see if the event CAN be enriched.  IIRC a p_log_type field is required on the event for it to be enriched. We can probably pre-populate that if we aren't already doing that (just gets a bit tricky with a rule that has multiple log types)

panther_analysis_tool/main.py
```diff
@@ -1098,6 +1100,104 @@ def classify_analysis(
     return all_specs, invalid_specs
 
 
+def enrich_test_data(backend: BackendClient, args: argparse.Namespace) -> Tuple[int, str]:
```
maxrichie5, line 1103: Can we put this in a new file in `panther_analysis_tool/command`? I would like to maintain the pattern of splitting up commands into their own files and putting them their instead of more stuff in main

panther_analysis_tool/main.py
```diff
@@ -1098,6 +1100,104 @@ def classify_analysis(
...
+
+            policy_id = item.analysis_spec.get("PolicyID", "")
+            if policy_id != "":
+                raw_analysis_items_by_id[policy_id] = item
+                continue
+
+            logging.warning(
+                "Analysis item %s is not a Rule, Scheduled Rule, or Policy - skipping",
+                item.spec_filename,
+            )
```
maxrichie5, line 1178: Might be better as an `info`

tests/unit/panther_analysis_tool/test_enriched_event_generator.py
```diff
@@ -0,0 +1,415 @@
...
+
+        enricher = EnrichedEventGenerator(backend=backend)
+
+        # The `enrich_test_data` method writes to the operating system so we'll
+        # mock `open` and assert on the content from the writes.
+        m = mock_open()
+        with patch('builtins.open', m):
+            enricher.enrich_test_data(test_data.values())
+
+        m().write.assert_has_calls([call('RuleID'),
```
maxrichie5, line 215: This is.... quite the test you got here

maxrichie5 (commented):
Nice!! Some asks on here

panther_analysis_tool/enriched_event_generator.py
```diff
@@ -0,0 +1,180 @@
...
+            for test in tests:
+                logging.info("\tEnriching test case '%s' for %s", test["Name"], analysis_id)
+                if "Log" in test:
+                    enriched_test = self._handle_rule_test(analysis_id, test)
+                    if enriched_test:
+                        enriched_tests.append(enriched_test)
+                elif "Resource" in test:
+                    enriched_test = self._handle_policy_test(analysis_id, test)
+                    if enriched_test:
+                        enriched_tests.append(enriched_test)
```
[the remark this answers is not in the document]
darwayne: yeah that's a bug in the UI if we're showing it there.

panther_analysis_tool/enriched_event_generator.py
darwayne: yeah we may need that then. The way you currently have the logic in place a test might get modified even if it wasn't enriched. Are you doing any special checks for that scenario?

panther_analysis_tool/enriched_event_generator.py
```diff
@@ -0,0 +1,180 @@
...
+        test[test_case_field_key]["p_enrichment"] = enriched_test_data.get("p_enrichment", {})
+
+        # Some test cases are pasted in as JSON. JSON does not roundtrip
+        # nicely - often just rendering as one giant line after we add
+        # the p_enrichment field.
+        #
+        # We're forcibly converting the test case data to a Python dict
+        # so that it roundtrips out as YAML instead. This is noisy
+        # for those tests that are in JSON format, but it's preferable
+        # to the alternative.
```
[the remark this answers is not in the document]
darwayne: ahh fun. Yeah that sounds like a pain

darwayne (commented):
After chatting about the intricacies of how this will work I'm fine with what we have. My biggest concern was that we'd be modifying test data unnecessarily but we have checks in place already to avoid that 👍🏾

darwayne (approved):
LGTM. CI isn't happy though

maxrichie5 (approved):

maxrichie5 (approved):

PR #352, opened 2023-08-22
pat cmd alias

maxrichie5:
### Background

Let's be honest, we all say and want to type PAT.

### Changes

* Adds `pat` as a cmd so users will be able to install then run `panther_analysis_tool` or `pat` (hopefully inside their pipenv)

### Testing

* Installed pat locally in panther-analysis and it ran `pipenv run pat -h` successfully


setup.py
```diff
@@ -1,56 +1,56 @@
...
-    keywords=['Security', 'CLI'],
-    scripts=['bin/panther_analysis_tool'],
+    license="AGPL-3.0",
+    description="Panther command line interface for writing, testing, and packaging policies/rules.",
+    author="Panther Labs Inc",
+    author_email="pypi@runpanther.io",
+    url="https://github.com/panther-labs/panther_analysis_tool",
+    download_url=f"https://github.com/panther-labs/panther_analysis_tool/archive/v{PAT_VERSION}.tar.gz",
+    keywords=["Security", "CLI"],
+    scripts=["bin/panther_analysis_tool", "bin/pat"],
```
maxrichie5, line 46: this is the real change. All else is formatting

dekatzenel (approved):
I was initially worried that this would override some other tool called "pat" that customers use but it looks like there isn't a common "pat" tool, so it's probably fine

PR #354, opened 2023-09-06
[DRAFT] Introduce obfuscation and deobfuscation commands

atbabers:
### Background

I've introduced two new commands to the panther_analysis_tool: obfuscation and deobfuscation. These features are inspired by a recently crafted tool, [logfuscate](https://github.com/atbabers/logfuscate). The primary objective is to mask sensitive details in unit tests (events) while at-rest in an organization's detection repository, recognizing that unit tests (events/logs) might hold confidential data.

### Changes

* Introduced a command to obfuscate detection metadata (.YML) files.
* Added a command to reverse obfuscation on affected files.
* Added custom key and pattern support.

### Considerations

* ~In line with logfuscate, there's may be a need to incorporate a consistent layer of obfuscation to ensure that the masked data remains alphanumeric.~
    * Migrated to base62 to ensure only alphanumeric characters are used in the encoding process.
* ~Based on previous discussions, it might be beneficial to separate obfuscate and deobfuscate functionalities into distinct .py files.~
    * I've decided against separating these functions into their own files. If it makes sense otherwise, please let me know.
* ~Exploring further optimization, especially for integration with GH Actions/CircleCI Orbs.~
    * For loops against files ending in .yml inside the rules folder should suffice.

### Testing

* Conducted Manual Testing
* Added some test cases


LucySuddenly (commented):
Hello @atbabers! Lucy from the Detections team here :) thank you for this really thoughtful contribution -- we definitely recognize the value of being able to obfuscate sensitive details which may need to be checked-in to version control. 

We're working on reorienting PAT to be more API-driven, and moving away from requirements like checking in sensitive data is part of the reason. We chatted this feature over as a team, and we agreed that with planned changes this would be deprecated too quickly.

Again, thank you for this contribution -- the consideration and rigor on display here is appreciated!

PR #361, opened 2023-09-19
deleting all things panther sdk

maxrichie5:
### Background

Execute order 66 (on the sdk because it is no longer used)

### Changes

* deleting anything relating to the sdk

### Testing

* ci?


panther_analysis_tool/backend/lambda_client.py
```diff
@@ -32,7 +31,6 @@
     BackendResponse,
     BulkUploadParams,
     BulkUploadResponse,
-    BulkUploadStatistics,
```
dekatzenel, line 35: Isn't this still used for regular bulk upload? Is that just not in this file?

dekatzenel (approved):

panther_analysis_tool/backend/lambda_client.py
maxrichie5: Good catch. I'll add that back

maxrichie5: Oh wait nevermind it was just no longer being used in this file. That class is still there

PR #368, opened 2023-09-27
Add backend check to `validate`

grantjoy:
### Background

Validate only works over API key. This helps users understand that the validate mechanism doesn't work in the way they are attempting to use it.

### Changes

* Adds a better error message when validate is called without an API key being set

### Testing

* manually tested `pat validate`


nskobov (approved):

panther_analysis_tool/util.py
```diff
@@ -140,6 +140,9 @@ def get_backend(args: argparse.Namespace) -> BackendClient:
             )
         )
 
+    if not hasattr(args, "aws_profile"):
+        return False
```
darwayne, line 144: hmm linter isn't happy about this since there's a type mismatch

stedrow, line 144:
```suggestion
        return 1, "--aws-profile wasn't set to a valid aws profile, please include a valid aws-profile in your arguments and try again."
```

grantjoy: thank you

grantjoy:
closing in favor of #380

PR #375, opened 2023-10-03
Detections Inheritance

grantjoy:
### Background

Adds support for derived detections.

### Testing

* Manual testing. Doesn't affect detections at all unless they have BaseDetection set, which would just be the new inheritance detections only.


panther_analysis_tool/schemas.py
```diff
@@ -154,44 +154,95 @@ def validate(
...
+    Optional("AlertTitle"): str,
+    Optional("AlertContext"): object,
+    Optional("GroupBy"): object,
+    Optional("BaseDetection"): And(str, NAME_ID_VALIDATION_REGEX),
+}
+
+RULE_SCHEMA = Schema(dict_schema, ignore_extra_keys=False)
+
+
+def validate_rule(data):
```
stedrow, line 241: is there ever invoked? I'm not seeing where we're actually running this

panther_analysis_tool/schemas.py
```diff
@@ -154,44 +154,95 @@ def validate(
...
+            "Enabled",
+            "Severity",
+            "Filename",
+            "Detection",
+            "LogTypes",
+            "ScheduledQueries",
+        }
+        for field in base_required_fields:
+            if field not in d:
+                raise SchemaError(f"'{field}' is required when 'BaseDetection' is not present.")
```
darwayne, line 175: I wouldn't even mention BaseDetection in this context.

grantjoy: thanks. removing all this.

panther_analysis_tool/schemas.py
```diff
@@ -193,6 +193,33 @@ def validate(
...
+        Optional("SummaryAttributes"): [str],
+        Optional("Threshold"): int,
+        Optional("Tags"): [str],
+        Optional("Reports"): {str: list},
+        Optional("DynamicSeverities"): object,
+        Optional("AlertTitle"): str,
+        Optional("AlertContext"): object,
+        Optional("GroupBy"): object,
+    },
+    ignore_extra_keys=False,
```
darwayne, line 220: I don't see this being used anywhere I think you'll need to update [this section](https://github.com/panther-labs/panther_analysis_tool/blob/6b45e21d132e3a0d59029d4509d6e920984c5a43/panther_analysis_tool/main.py#L2030) to include the derived schema and set to true conditionally as we seem to do for other types

panther_analysis_tool/schemas.py
```diff
@@ -193,6 +193,33 @@ def validate(
     ignore_extra_keys=False,
 )  # Prevent user typos on optional fields
 
+DERIVED_SCHEMA = Schema(
+    {
+        "AnalysisType": Or("rule", "scheduled_rule"),
```
darwayne, line 198: this should just be rule for now

darwayne (commented):
Nice stuff, think we're almost there

panther_analysis_tool/schemas.py
grantjoy: Added, thanks.

panther_analysis_tool/schemas.py
grantjoy: changed, thanks.

darwayne (approved):

PR #380, opened 2023-10-16
Check that backend was set successfully

grantjoy:
### Background

It's not clear that an API token is necessary to run `validate` and `benchmark` commands. This PR improves the error messaging.

### Changes
* throw an error and catch it in a sane way when backend is not set correctly
* makes sure some type of backend is always set
* makes sure validate and benchmark are set using API key6s

### Testing

* manually tested


panther_analysis_tool/command/benchmark.py
```diff
@@ -35,8 +35,8 @@ def __str__(self) -> str:
 def run(  # pylint: disable=too-many-locals
     backend: BackendClient, args: argparse.Namespace
 ) -> Tuple[int, str]:
-    if not backend.supports_perf_test():
-        return 1, "benchmark is only supported via the api token"
+    if backend is None or not backend.supports_perf_test():
+        return 1, "Invalid backend: `benchmark` is only supported via API token"
```
nskobov, line 39: nit: `Invalid backend: benchmark` to me reads like "benchmark" is the name of the backend. I'd use `.` instead of `:`

nskobov (dismissed):

panther_analysis_tool/command/benchmark.py
grantjoy, line 39:
```suggestion
        return 1, "Invalid backend. `benchmark` is only supported via API token"
```

panther_analysis_tool/command/validate.py
```diff
@@ -15,8 +15,8 @@
 
 
 def run(backend: BackendClient, args: argparse.Namespace) -> Tuple[int, str]:
-    if not backend.supports_bulk_validate():
-        return 1, "bulk validate is only supported via the api token"
+    if backend is None or not backend.supports_bulk_validate():
+        return 1, "Invalid backend: `validate` is only supported via API token"
```
grantjoy, line 19:
```suggestion
        return 1, "Invalid backend. `validate` is only supported via API token"
```

nskobov (approved):

PR #383, opened 2023-10-18
bug fix for zip_chunker

grantjoy:
### Background

introduced a bug to zip chunker yesterday, this attempts to solve it

### Changes

*  include the necessary `not` for `if analysis_spec is not None and "Filename" in analysis_spec:`

### Testing

* manual


darwayne (approved):

panther_analysis_tool/zip_chunker.py
```diff
@@ -186,7 +186,7 @@ def chunk_analysis(
         for chunk in chunk_files:
             if chunk.matches_file(analysis_spec_filename, analysis_spec):
                 chunk.add_file(main_file)
-                if analysis_spec is None and "Filename" in analysis_spec:
+                if "Filename" in analysis_spec:
```
grantjoy:
```suggestion
                if analysis_spec is not None and "Filename" in analysis_spec:
```

darwayne (approved):

PR #389, opened 2023-10-20
README makeover

grantjoy:
Rewriting the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,73 +1,74 @@
-<h1 align="center">Panther Analysis Tool</h1>
+# Panther Analysis Tool
+*Command Line Panther Analysis Management*
 
-<p align="center">
-  <i>Command Line Panther Analysis Management</i>
-</p>
-
-<p align="center">
-  <a href="https://docs.runpanther.io">Documentation</a> |
-  <a href="https://docs.runpanther.io/quick-start">Quick Start</a>
-</p>
-
----
-
-`panther_analysis_tool` is a Python CLI for testing, packaging, and deploying Panther Detections defined in code. See the [Panther documentation](https://docs.runpanther.io/quick-start) for more details on Panther.
+[Panther Analysis Tool](https://github.com/panther-labs/panther_analysis_tool) is a Python application for testing, packaging, and deploying Panther Detections. For further details, see [Quick Start](https://docs.runpanther.io/quick-start) and [Panther Documentation](https://docs.panther.com/).
 
 # Installation
 
-Install simply with pip:
+## From PyPi
 
-```bash
+Use pip to install [panther_analysis_tool package](https://pypi.org/project/panther-analysis-tool/) from PyPi:
+
+```shell
 pip3 install panther_analysis_tool
 ```
 
-If you would rather use the `panther_analysis_tool` outside of the virtual environment, install it  directly:
+Or without a virtual environment:
 
-```bash
+```shell
 make deps
 pip3 install -e .
 ```
 
-## Build From Source
+Or from source:
 
-If you'd prefer instead to run from source for development reasons, first setup your environment:
-
-```bash
+```shell
 make install
 pipenv run -- pip3 install -e .
 ```
 
-# Commands and Usage
+# Usage
 
-## View available commands:
+## Help
 
-```bash
+Show available commands and their options:
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,11 +95,11 @@ AWS.IAM.MFAEnabled
 [INFO]: <current working directory>/tmp/panther-analysis-2020-03-23T12-48-18.zip
 ```
 
-### Upload packages to Panther directly. 
+Or upload packages directly into Panther:
 
 _Note, this expects your environment to be setup the same way as if you were using the AWS CLI, see the setup instructions [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html). We also recommend using a credentials manager such as [aws-vault](https://github.com/99designs/aws-vault)._
```
grantjoy, line 100: Is this still true? Can't we use API keys OR AWS creds?

nskobov (approved):
looks good to me. for other reviewers i found it helpful to look at the rendered view instead of the text diff:
https://github.com/panther-labs/panther_analysis_tool/blob/d617a4f0a20d22804e2f292138c2343379ce59d7/README.md

README.md
darwayne: correct users can utilize an api key OR aws creds, probably be good to call that out I don't think we ever updated readme to reflect that

darwayne (dismissed):

README.md
grantjoy: Thanks @darwayne can you stamp again? I removed that line.

README.md
```diff
@@ -125,20 +127,20 @@ AWS.CloudTrail.MFAEnabled
   "totalRules": 1
 }
 ```
-## Filtering
+### Filtering
 
-The `test`, `zip`, and `upload` commands all support filtering. Filtering works by passing the `--filter` argument with a list of filters specified in the format `KEY=VALUE1,VALUE2`. The keys can be any valid field in a policy or rule. When using a filter, only anaylsis that matches each filter specified will be considered. For example, the following command will test only items with the AnalysisType of policy AND the severity of High:
+The `test`, `zip`, and `upload` commands all support filtering. Filtering works by passing the `--filter` argument with a list of filters specified in the format `KEY=VALUE1,VALUE2`. The keys can be any valid field in a policy or rule. When using a filter, only anaylsis that matches each filter specified will be considered. For example, the following command will test only items with the AnalysisType as `policy` AND severity as `High``:
```
grantjoy, line 132:
```suggestion
The `test`, `zip`, and `upload` commands all support filtering. Filtering works by passing the `--filter` argument with a list of filters specified in the format `KEY=VALUE1,VALUE2`. The keys can be any valid field in a policy or rule. When using a filter, only anaylsis that matches each filter specified will be considered. For example, the following command will test only items with the AnalysisType as `policy` AND severity as `High`:
```

README.md
```diff
@@ -125,20 +127,20 @@ AWS.CloudTrail.MFAEnabled
...
+$ panther_analysis_tool test --path tests/fixtures/valid_policies --filter AnalysisType=policy Severity=High
 [INFO]: Testing analysis packs in tests/fixtures/valid_policies
 
 AWS.IAM.BetaTest
 	[PASS] Root MFA not enabled fails compliance
 	[PASS] User MFA not enabled fails compliance
 ```
 
-Whereas the following command will test items with the AnalysisType policy OR rule, AND the severity High:
+Alternately, the following command will test items with the AnalysisType `policy` OR `rule``, AND the severity `High``:
```
grantjoy, line 143:
```suggestion
Alternately, the following command will test items with the AnalysisType `policy` OR `rule`, AND the severity `High`:
```

README.md
```diff
@@ -217,48 +219,55 @@ Failed Tests Summary
...
+From your (panther_analysis)[https://github.com/panther-labs/panther-analysis] content repository, run:
+```shell
 pipenv install --editable ../relative/path/to/panther_analysis_tool
 ```
 
-Then you can test your changes to PAT by running 
-```
+Then test changes to PAT:
+```shell
 pipenv run panther_analysis_tool <cmd>
```
grantjoy, line 271:
```suggestion
pipenv run panther_analysis_tool test
```

nskobov (approved):

PR #405, opened 2023-11-02
Add additional CarbonBlack log type regexes

egibs:
### Background

In order to author new CarbonBlack rules, we need PAT to support the base log type regex patterns.

### Changes

- Adds `CarbonBlack.AlertV2`
- Adds `CarbonBlack.EndpointEvent`
- Adds `CarbonBlack.WatchlistHit`

### Testing

* Tests pass as expected


panther_analysis_tool/schema_regexs.py
```diff
@@ -31,6 +31,9 @@
     r"Bitwarden\.Events|"
     r"Box\.Event|"
     r"CarbonBlack\.Audit|"
+    r"CarbonBlack\.AlertV2|"
+    r"CarbonBlack\.EndpointEvent|"
+    r"CarbonBlack\.WatchlistHit|"
```
nskobov, line 36: was this added manually? my understanding is we have automation that generates this file. did it not run somehow?

egibs: Ah, yea -- `# Autogenerated using the pat_regex_generator dev tool`. I'll run that instead.

PR #410, opened 2023-11-09
Whitelisting 'tests' as an allowed field in derived detection YAML

yusufak-panther:
### Background

We had previously disallowed defining tests in a derived detection YAML. This change reverts that decision as defining tests is a safe operation, even if inheritance is not supported for tests right now.

### Changes

* Added `Tests` to the Schema governing the list of valid keys in a derived detection YAML

### Testing

* Manually tested uploading a derived detection with tests


LucySuddenly (approved):
does what it says on the tin!

grantjoy (approved):

yusufak-panther:
Closing and re-opening due to issue with git config

PR #411, opened 2023-11-13
Add CORRELATION_RULE AnalysisType.

zacbrown:
### Background

Add new AnalysisType for Correlation Rules. This does not include any work around enabling testing or testing uploading yet. That relies on landing backend code that cannot yet be tested against.

### Changes

Add new AnalysisType for Correlation Rules.

### Testing

* updated existing tests
* manual testing w/ a pseudo-`correlation_rule` to ensure no crashing


darwayne:
since tests aren't supported should we exclude that from our local testing logic as well?

dekatzenel (approved):
This seems fine as far as it goes

PR #416, opened 2023-11-29
Add schema support for Correlation Rules

zacbrown:
### Background

This change adds schemas for Correlation Rules and explicitly skips testing them when `panther_analysis_tool test` is called.

### Changes

- Add schemas for correlation_rule
- Skip running tests on correlation_rule
- Log message when skipping correlation_rule tests.

### Testing

* local testing, existing tests


panther_analysis_tool/main.py
```diff
@@ -911,6 +912,12 @@ def setup_run_tests(  # pylint: disable=too-many-locals,too-many-arguments
             filters=analysis_spec.get(BACKEND_FILTERS_ANALYSIS_SPEC_KEY) or None,
         )
 
+        if is_correlation_rule(analysis_spec):
+            logging.warning(
+                "Skipping Correlation Rule '%s', testing not supported", analysis_spec.get("RuleID")
+            )
```
maxrichie5, line 918: I feel like we don't need a message here. If testing isn't supported, it is verbose to state that every time tests are run. We can make this a debug instead so that info is still available

sesheffield (approved):

panther_analysis_tool/main.py
maxrichie5: I think we should if they are normally something that is tested, but this is a special case where are not tested. But testing for correlation rules isn't even a part of our product (so far). You don't think that would clutter up user's output when running tests?

PR #419, opened 2023-12-06
conditionally show correlation rule output

maxrichie5:
### Background

<High level overview here>

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


panther_analysis_tool/main.py
```diff
@@ -369,7 +374,14 @@ def upload_zip(
                 else:
                     response = backend.bulk_upload(upload_params)
 
-                logging.info("API Response:\n%s", json.dumps(asdict(response.data), indent=4))
+                resp_dict = asdict(response.data)
+                flags_params = FeatureFlagsParams(
+                    flags=[FeatureFlagWithDefault(flag=ENABLE_CORRELATION_RULES_FLAG)]
```
nskobov, line 379: what does ENABLE_CORRELATION_RULES_FLAG determine?

nskobov (approved):
lgtm 🚀

panther_analysis_tool/main.py
maxrichie5: That is a constant set to the name of the flag in split. The backend evaluates what the value of the flag is set to and returns true if it is on and false otherwise

PR #422, opened 2023-12-15
signal support

maxrichie5:
### Background

Adding support for signals

### Changes

* Signals can be uploaded and results appear in output (only when FF is on)
* Signals can be tested
* Signals can be filtered
* Parsing filters was moved to another file to be more organized. Lots of test changes as a result

### Testing

* unit tests added and all pass
* manually tested filtering 
* manually tested testing
* manually tested upload

### DIFF
--- a/panther_analysis_tool/analysis_utils.py
+++ b/panther_analysis_tool/analysis_utils.py
@@ -46,6 +46,7 @@ from panther_analysis_tool.constants import (
     POLICIES_PATH_PATTERN,
     QUERIES_PATH_PATTERN,
     RULES_PATH_PATTERN,
+    SIGNALS_PATH_PATTERN,
     VERSION_STRING,
     AnalysisTypes,
 )
@@ -425,6 +426,7 @@ def load_analysis_specs_ex(
                             HELPERS_PATH_PATTERN,
                             LUTS_PATH_PATTERN,
                             RULES_PATH_PATTERN,
+                            SIGNALS_PATH_PATTERN,
                             PACKS_PATH_PATTERN,
                             POLICIES_PATH_PATTERN,
                             QUERIES_PATH_PATTERN,
--- a/panther_analysis_tool/analysis_utils.py
+++ b/panther_analysis_tool/analysis_utils.py
@@ -20,9 +20,11 @@ import dataclasses
 import json
 import logging
 import os
+from distutils.util import strtobool  # pylint: disable=E0611, E0401
 from fnmatch import fnmatch
 from typing import Any, Callable, Dict, Generator, Iterator, List, Optional, Tuple
 
+import schema
 from ruamel.yaml import YAML
 from ruamel.yaml import parser as YAMLParser
 from ruamel.yaml import scanner as YAMLScanner
@@ -46,7 +48,24 @@ from panther_analysis_tool.constants import (
     VERSION_STRING,
     AnalysisTypes,
 )
+from panther_analysis_tool.schemas import (
+    GLOBAL_SCHEMA,
+    POLICY_SCHEMA,
+    RULE_SCHEMA,
+    CORRELATION_RULE_SCHEMA,
+    SIGNAL_SCHEMA,
+    DERIVED_SCHEMA,
+)
 from panther_analysis_tool.util import is_simple_detection
+
+_VALID_FILTER_KEYS = (
+    list(GLOBAL_SCHEMA.schema.keys())
+    + list(POLICY_SCHEMA.schema.keys())
+    + list(RULE_SCHEMA.schema.keys())
+    + list(CORRELATION_RULE_SCHEMA.schema.keys())
+    + list(SIGNAL_SCHEMA.schema.keys())
+    + list(DERIVED_SCHEMA.schema.keys())
+)
 
 
 class ClassifiedAnalysis:
@@ -129,10 +148,57 @@ class ClassifiedAnalysisContainer:
             self.queries.append(classified_analysis)
 
 
+@dataclasses.dataclass
+class AnalysisFilters:
+    filters: Dict[str, List]
+    filters_inverted: Dict[str, List]
+
+    @staticmethod
+    def from_filter_list(filters: List[str]) -> "AnalysisFilters":
+        parsed_filters: Dict[str, List[Any]] = {}
+        parsed_filters_inverted: Dict[str, List[Any]] = {}
+
+        for filt in filters:
+            split = filt.split("=")
+            if len(split) != 2 or split[0] == "" or split[1] == "":
+                logging.warning("Filter %s is not in format KEY=VALUE, skipping", filt)
+                continue
+            # Check for "!="
+            invert_filter = split[0].endswith("!")
+            if invert_filter:
+                split[0] = split[0][:-1]  # Remove the trailing "!"
+            key = split[0]
+
+            if key not in _VALID_FILTER_KEYS and schema.Optional(key) not in _VALID_FILTER_KEYS:
+                logging.warning("Filter key %s is not a valid filter field, skipping", key)
+                continue
+
+            if invert_filter:
+                parsed_filters_inverted[key] = split[1].split(",")
+            else:
+                parsed_filters[key] = split[1].split(",")
+
+            # Handle boolean fields
+            if key == "Enabled":
+                try:
+                    bool_value = bool(strtobool(split[1]))
+                except ValueError:
+                    logging.warning("Filter key %s should have either true or false, skipping", key)
+                    continue
+                if invert_filter:
+                    parsed_filters_inverted[key] = [bool_value]
+                else:
+                    parsed_filters[key] = [bool_value]
+        return AnalysisFilters(parsed_filters, parsed_filters_inverted)
+
+    def empty(self) -> bool:
+        return len(self.filters) == 0 and len(self.filters_inverted) == 0
+
+
 def filter_analysis(
-    analysis: List[ClassifiedAnalysis], filters: Dict[str, List], filters_inverted: Dict[str, List]
+    analysis: List[ClassifiedAnalysis], analysis_filters: AnalysisFilters
 ) -> List[ClassifiedAnalysis]:
-    if filters is None:
+    if analysis_filters.empty():
         return analysis
 
     filtered_analysis = []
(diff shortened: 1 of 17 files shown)


panther_analysis_tool/analysis_utils.py
```diff
@@ -129,10 +150,61 @@ def add_classified_analysis(
             self.queries.append(classified_analysis)
 
 
+@dataclasses.dataclass
+class AnalysisFilters:
+    filters: Dict[str, List]
+    filters_inverted: Dict[str, List]
+
+    @staticmethod
+    def from_filter_list(filters: List[str]) -> "AnalysisFilters":
```
maxrichie5, line 159: This whole function was just copied from [parse_filters in main.py](https://github.com/panther-labs/panther_analysis_tool/pull/422/files#diff-4c3406e46ca13890fa64d295f86ed01e0b0bcd3d28031f1fc97978bb8fca2da3L2015) and I added a couple little changes

darwayne (dismissed):

panther_analysis_tool/analysis_utils.py
```diff
@@ -129,10 +150,61 @@ def add_classified_analysis(
             self.queries.append(classified_analysis)
 
 
+@dataclasses.dataclass
+class AnalysisFilters:
```
broox, line 154: 🙏

panther_analysis_tool/analysis_utils.py
broox: aside from switching this up to be a class, the main logic looks to be the addition of new filter keys, which lgtm.

that said, unit testing this class and its instance methods might be nice.

panther_analysis_tool/backend/client.py
```diff
@@ -233,6 +233,7 @@ def is_valid(self) -> bool:
         return True
 
 
+# pylint: disable=too-many-instance-attributes
```
broox, line 236: 🙀

panther_analysis_tool/errors.py
```diff
@@ -0,0 +1,2 @@
+class UnknownAnalysisIdError(Exception):
+    pass
```
broox, line 2: 🙏

panther_analysis_tool/main.py
```diff
@@ -380,6 +382,7 @@ def upload_zip(
                 )
                 if not backend.feature_flags(flags_params).data.flags[0].treatment:
```
broox, line 383: nit unrelated to these changes: without expanding the code above, it wasn't quite clear what this line of code was doing... particularly what `flags_params` meant.

it might be more clear if we renamed `flags_params` to `correlation_rule_flags`... or better yet, we could create a helper function that wrapped this logic to make the whole thing easier to understand and also re-use.

e.g.
```
def is_feature_enabled(flag_name: str) -> bool:
    flags_params = FeatureFlagsParams(
        flags=[FeatureFlagWithDefault(flag=flag_name)]
    )
    return backend.feature_flags(flags_params).data.flags[0].treatment
```

and then:
```
if not is_feature_enabled(ENABLE_CORRELATION_RULES_FLAG):
    del resp_dict["correlation_rules"]
    del resp_dict["signals"]
```

panther_analysis_tool/main.py
```diff
@@ -925,7 +923,7 @@ def setup_run_tests(  # pylint: disable=too-many-locals,too-many-arguments,too-m
         analysis_type = analysis_spec["AnalysisType"]
 
         detection_args = dict(
-            id=analysis_spec.get("PolicyID") or analysis_spec["RuleID"],
+            id=lookup_analysis_id(analysis_spec),
```
broox, line 926: 🙏

panther_analysis_tool/main.py
```diff
@@ -1087,7 +1085,7 @@ def classify_analysis(
             analysis_schema.validate(analysis_spec)
 
             # lookup the analysis type id and validate there aren't any conflicts
-            analysis_id = lookup_analysis_id(analysis_spec, analysis_type)
+            analysis_id = lookup_analysis_id(analysis_spec)
```
broox, line 1088: 😄

panther_analysis_tool/main.py
```diff
@@ -1204,32 +1202,23 @@ def enrich_test_data(backend: BackendClient, args: argparse.Namespace) -> Tuple[
...
             1,
-            f"No analysis content in {args.path} matched filters {args.filter} - {args.filter_inverted}",
+            f"No analysis content in {args.path} matched filters {args.filter}",
         )
 
     # We only used classify_analysis to figure out what to filter out, if anything.
     # We need to filter our own list of items now, using what's left in the `specs` variable.
     raw_analysis_items_by_id = {}
     for item in raw_analysis_items:
         if item.analysis_spec is not None:
```
broox, line 1218: nit: to avoid deep indentation, this could be:

```
if item.analysis_spec is None:
    continue
```

panther_analysis_tool/main.py
```diff
@@ -1204,32 +1202,23 @@ def enrich_test_data(backend: BackendClient, args: argparse.Namespace) -> Tuple[
...
-                raw_analysis_items_by_id[rule_id] = item
-                continue
-
-            policy_id = item.analysis_spec.get("PolicyID", "")
-            if policy_id != "":
-                raw_analysis_items_by_id[policy_id] = item
-                continue
+            analysis_id = lookup_analysis_id(item.analysis_spec)
+            if analysis_id != UNKNOWN_ANALYSIS_ID:
+                raw_analysis_items_by_id[analysis_id] = item
```
broox, line 1221: we either need a `continue` here, or a slight refactor, ortherwise the below info line will be logged on each loop.

a suggestion:
```
    analysis_id = lookup_analysis_id(item.analysis_spec)
    if analysis_id == UNKNOWN_ANALYSIS_ID:
        logging.info(
            "Analysis item %s is not a Rule, Scheduled Rule, or Policy - skipping",
            item.spec_filename,
        )
        continue

    raw_analysis_items_by_id[analysis_id] = item
```

panther_analysis_tool/main.py
```diff
@@ -1240,15 +1229,9 @@ def enrich_test_data(backend: BackendClient, args: argparse.Namespace) -> Tuple[
...
-            filtered_raw_analysis_items_by_id[rule_id] = raw_analysis_items_by_id[rule_id]
-            continue
-
-        policy_id = spec.analysis_spec.get("PolicyID", "")
-        if policy_id != "":
-            filtered_raw_analysis_items_by_id[policy_id] = raw_analysis_items_by_id[policy_id]
-            continue
+        analysis_id = lookup_analysis_id(spec.analysis_spec)
+        if analysis_id != UNKNOWN_ANALYSIS_ID:
+            filtered_raw_analysis_items_by_id[analysis_id] = raw_analysis_items_by_id[analysis_id]
```
broox, line 1234: similar to above, do we want to log an info line here if the `analysis_id` is unknown?

panther_analysis_tool/util.py
```diff
@@ -252,3 +253,31 @@ def log_and_write_to_file(msgs: List[str], filename: TextIO) -> None:
     for msg in msgs:
         filename.write(msg + "\n")
         logging.info(msg)
+
+
+def lookup_analysis_id(analysis_spec: Any) -> str:
+    analysis_type = analysis_spec.get("AnalysisType")
```
broox, line 259: good change 😄 
unit tests would be cool to introduce here.

panther_analysis_tool/validation.py
```diff
@@ -101,15 +103,9 @@ def validate_packs(analysis_specs: ClassifiedAnalysisContainer) -> List[Any]:
...
-            or analysis_spec.get("RuleID")
-            or analysis_spec.get("DataModelID")
-            or analysis_spec.get("GlobalID")
-            or analysis_spec.get("PackID")
-            or analysis_spec.get("QueryName")
-            or analysis_spec["LookupName"]
-        )
+        analysis_id = lookup_analysis_id(analysis_spec)
+        if analysis_id == UNKNOWN_ANALYSIS_ID:
+            raise UnknownAnalysisIdError(f"Unknown analysis id from analysis item {analysis_spec}")
```
broox, line 108: does this need to be caught anywhere?

broox (commented):
there's some really good cleanup in here, nice work!
a few more unit tests might be cool, otherwise, i just left a couple small suggestions and questions. nothing blocking.

it'd be good to have someone from the DAC team take a look since they own PAT?

panther_analysis_tool/util.py
```diff
@@ -252,3 +253,31 @@ def log_and_write_to_file(msgs: List[str], filename: TextIO) -> None:
...
+        analysis_id = analysis_spec["QueryName"]
+    elif analysis_type == AnalysisTypes.SAVED_QUERY:
+        analysis_id = analysis_spec["QueryName"]
+    elif analysis_type == AnalysisTypes.SIGNAL:
+        analysis_id = analysis_spec["SignalID"]
+    elif analysis_type in [
+        AnalysisTypes.RULE,
+        AnalysisTypes.SCHEDULED_RULE,
+        AnalysisTypes.CORRELATION_RULE,
+    ]:
```
broox, line 281: oh, also... i realize that this is how the function already behaved, so i don't wanna derail anything here... but i assume that we don't actually do anything with `UNKNOWN_ANALYSIS_ID`.

in the future, it might be better to add:
```
else:
    raise UnknownAnalyasisID
```

and then callers can handle that exception rather than check strings.

panther_analysis_tool/main.py
maxrichie5: I like that. I will switch it

panther_analysis_tool/validation.py
maxrichie5: I don't expect this to happen. But in the off chance it does, it will be caught at the highest level and be printed as an error

panther_analysis_tool/main.py
maxrichie5: Updated

panther_analysis_tool/main.py
maxrichie5: Updated

panther_analysis_tool/util.py
maxrichie5: Added some unit tests

PR #433, opened 2024-01-25
updates mock values to always process as a string to align with enterprise behavior.

darwayne:
### Background
Today if a mock is set as a bool value PAT treats the mock as a bool. But when we upload the detection to enterprise it gets converted to a string, which can create conflicting unit test results between local PAT and panther enterprise.

### Changes

* Updates PAT mocking logic to treat bools as a string the same way enterprise does today

### Testing
* Manual testing was done to verify this behavior


darwayne:
**Note** 
We have a few detections in panther-analysis that utilize a bool in the mocks, so until that is addressed we will more than likely see failed CI tests

darwayne:
Closing this out for now. We will solve this with a planned improvement to mocks

PR #448, opened 2024-02-08
Omit rules with Configuration Required Tag from Pack check

egibs:
### Background

We need to ignore Rules that contain the `Configuration Required` from Pack checks. This will also require that Packs containing these rules already be updated to not include these rules.

### Changes

* Ignores Rules that contain the `Configuration Required` Tag

### Testing

* Tests pass as expected

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -1316,7 +1316,9 @@ def check_packs(args: argparse.Namespace) -> Tuple[int, str]:
             if is_simple_pack != is_simple_rule:
                 # simple rules should be in simple packs
                 continue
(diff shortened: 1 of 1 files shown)


darwayne:
Lets avoid the version bumps for now in the pipfile

arielkr256 (approved):

darwayne (changes requested):
lets keep the changes to just logic changes. Want to avoid pipfile changes

egibs:
> lets keep the changes to just logic changes. Want to avoid pipfile changes

Addressed in [`1188350` (#448)](https://github.com/panther-labs/panther_analysis_tool/pull/448/commits/1188350d2633c8f2c63c5a166506d91126217756).

darwayne (approved):

PR #450, opened 2024-02-12
adding correlation rules to async bulk upload graphql

maxrichie5:
### Background

Correlation rule statistics were being returned from bulk upload, but the data wasn't actually being populated. 

### Changes

* Added correlation rules to both types of bulk upload

### Testing

* Manually tested

### DIFF
--- a/panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
+++ b/panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
@@ -21,12 +21,15 @@ query status($input: ID!) {
             queries {
                 ...UploadStatisticDetails
             }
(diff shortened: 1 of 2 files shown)


panther_analysis_tool/backend/graphql/async_bulk_upload_status.graphql
```diff
@@ -21,12 +21,15 @@ query status($input: ID!) {
             queries {
                 ...UploadStatisticDetails
             }
+            correlationRules {
+                ...UploadStatisticDetails
+            }
```
maxrichie5, line 26: I only added these chunks. The rest were the IDE autoformatting

nskobov (approved):

PR #459, opened 2024-02-22
Ignore disabled rules during packs-check

melenevskyi:
### Background

Disabled rules do not have to be in any pack. This PR updates `check-packs` to ignore disabled rules.


egibs (approved):

darwayne (changes requested):
lets move convo to our slack

PR #487, opened 2024-04-23
adds logic to improve temp directory management

darwayne:
### Background
If multiple instances of PAT are invoked users can see unexpected behavior as multiple processes can share the same paths causing various race conditions. 

### Changes
  - each instance of PAT gets its own temp folder
  - at process exit we remove the directory

### Testing

* manually tested running the following commands in parallel
   * `pipenv run pat upload --skip-tests`
   * `pipenv run pat upload`
   * `pipenv run pat test`

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -17,6 +17,7 @@ You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
+# pylint: disable=wrong-import-order, wrong-import-position, ungrouped-imports
 import argparse
 import base64
 import contextlib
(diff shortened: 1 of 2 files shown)


panther_analysis_tool/main.py
```diff
@@ -48,6 +49,11 @@
 from unittest.mock import MagicMock, patch
 from uuid import uuid4
 
+from panther_analysis_tool.directory import setup_temp
+
+# this is needed at this location so each process can have its own temp directory
+setup_temp()
```
nskobov, line 55: any way this can be opt in? I can imagine scenarios where PAT is running in an environment where users dont have access to create a temp directory, and this could break potentially

darwayne: We are already using the system temp directory today in other spots, this just creates a subdirectory there per process. 

So if users had issues with temp file handling they would've seen errors prior to this change

nskobov (approved):

PR #355, opened 2023-09-08
Update PAT to use Python3.11

egibs:
### Background

This PR updates PAT to use Python3.11 and also updates most of the outdated dependencies.

Additionally, this PR moves away from `nose` in favor of `nose2` to avoid a large lift re: tests.

### Changes

* Updates `python_version` to `3.11`
* Updates outdated dependenices
* Moves from `nose` to `nose2`

### Testing

* Tests pass as expected

PR #492, opened 2024-04-25
bump python to 3.11 in order to bump vulnerable aiohttp

le4ker:
### Background

Bumps Python, in order to [bump aiohttp through the gql package](https://github.com/graphql-python/gql/blob/23636985857affd9b35bfc895f4bafdf2dc0801c/setup.py#L43).

This fixes the following vulnerabilities:r

### Changes

* <Describe changes here>

### Testing

* <Testing steps>

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -50,4 +50,4 @@ jsonlines = "==3.0.0"
 panther-core = "==0.8.1"
 
 [requires]
-python_version = "3.9"
+python_version = "3.11"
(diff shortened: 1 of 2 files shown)


.github/workflows/fmt.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.11
```
egibs, line 19: This may have other implications.

Here's my previous PR: https://github.com/panther-labs/panther_analysis_tool/pull/355

le4ker: That's great! I'll close this one 😄

PR #493, opened 2024-04-25
Update PAT to use Python3.11

egibs:
### Background

This PR upgrades from Python3.9 to 3.11.

### Changes

* Upgrades Python from 3.9 to 3.11
* Updates any failing tests/checks where appropriate

### Testing

* Tests/Checks pass as expected


panther_analysis_tool/backend/public_api_client.py
```diff
@@ -154,7 +153,9 @@ def feature_flags_query(self) -> DocumentNode:
 
     def _load(self, name: str) -> DocumentNode:
         if name not in self._cache:
-            self._cache[name] = Path(_get_graphql_content_filepath(name)).read_text()
+            file_path = _get_graphql_content_filepath(name)
```
dekatzenel, line 156: Can you elaborate on this change?

panther_analysis_tool/main.py
```diff
@@ -580,7 +575,7 @@ def publish_release(args: argparse.Namespace) -> Tuple[int, str]:
         "Authorization": f"token {api_token}",
     }
     # check this tag doesn't already exist
-    response = requests.get(release_url + f"/tags/{args.github_tag}", headers=headers)
+    response = requests.get(release_url + f"/tags/{args.github_tag}", headers=headers, timeout=10)
```
dekatzenel, line 578: How do we know that 10 is the right value here?

tests/unit/panther_analysis_tool/test_enriched_event_generator.py
```diff
@@ -25,7 +23,6 @@
 )
 
 
-@nottest
```
dekatzenel, line 28: Is this not needed now?

dekatzenel (commented):
A few questions

panther_analysis_tool/backend/public_api_client.py
egibs: I thought I needed `with open` to specify the file encoding; turns out that's not the case.

Reverted and updated 👍🏻

panther_analysis_tool/main.py
egibs: It's just an arbitrary value. Five would probably suffice but either should be sufficient.

tests/unit/panther_analysis_tool/test_enriched_event_generator.py
egibs: Correct, `nose2` doesn't have that decorator: https://docs.nose2.io/en/latest/decorators.html

dekatzenel: That decorator was used to tell nose that function isn't a test a shouldn't be treated as such. How do we replicate that functionality in nose2?

egibs: Looks like we can use `@unittest.skip` -- I'll push that change.

dekatzenel (approved):

tests/unit/panther_analysis_tool/test_enriched_event_generator.py
egibs: I moved the code to its own module in [`8cf5785` (#493)](https://github.com/panther-labs/panther_analysis_tool/pull/493/commits/8cf5785538d8d2a5f2d9347c25b9c8b731d744e6). 👍🏻

dekatzenel (approved):

PR #503, opened 2024-05-07
Ignore dependabot.yml

egibs:
### Background

We have a pattern for ignoring files and this needs to be extended to include `dependabot.yml` since PAT treats any JSON or YAML files as valid files.

### Changes

* Adds `dependabot.yml` to `ignored_normalized`

### Testing

* Checks pass as expected


arielkr256:
Makes ignored_normalized a set - I don't see this

egibs:
> Makes ignored_normalized a set - I don't see this

I reverted that change -- updated the PR to reflect that.

PR #508, opened 2024-05-30
Move panther-core from dev-packages to packages

egibs:
### Background

We're seeing an interesting anomaly when installing PAT in `panther-analysis` where `panther-core` `0.10.0` will be uninstalled in favor of `0.8.1` even though we specify the former in `Pipfile.lock`.

This PR moves the `panther-core` package to the `packages` section of `Pipfile` so that we always have it pinned in both `Pipfile` and `requirements.txt` (via `make deps-update`).

### Changes

* Moves `panther-core` to `[packages]`
* Runs `make deps-update`

### Testing

* Checks pass as expected


requirements.txt
```diff
@@ -32,18 +32,19 @@ markupsafe==2.1.5; python_version >= '3.7'
 multidict==6.0.5; python_version >= '3.7'
 nested-lookup==0.2.25
 packaging==23.1; python_version >= '3.7'
+panther-core==0.10.0
```
egibs, line 35: Hopefully this resolves the issues we're seeing downstream.

arielkr256 (approved):

PR #510, opened 2024-06-03
Remove panther-core pin in setup.py, update to 0.10.1

egibs:
### Background

@le4ker noticed that `panther-core` was pinned to `0.8.1` in `setup.py`. This is the likely cause of the downgrade behavior we were seeing in PA.

### Changes

* Removes `panther-core` pin in `setup.py`
* Updates `panther-core` to `0.10.1`

### Testing

* Tests pass as expected


arielkr256 (approved):
Nice catch!

PR #532, opened 2024-08-14
removes test command from ci job

darwayne:
### Background

We recently started seeing failures when running the make test command, even when navigating back to old commits. This is likely due to some runtime change on the github action side. 

While this removes the test command it retains the integration command which continues to work; a deeper investigation is needed to understand why make test has recently started failing.

This change makes it so we can unblock other PRs and still have reasonable confidence in our changes since we'll still have integration tests running

### Changes

* Removes the test command from the ci command

### Testing
* Lots of manual testing locally and via github action


dekatzenel (approved):
TIHI

PR #536, opened 2024-08-27
updates release job to run integration test instead of unit tests

darwayne:
### Background
We still need to get to the bottom of the unit test issue we've been seeing. In the mean time we'll update our release job to run integration tests instead.

### Changes

* updates release job to run integration tests

### Testing

* [x] manually verify that build step with just the integration logic will work


LucySuddenly (approved):
does what it says on the tin!

PR #539, opened 2024-09-09
PAT upload: sync capabilities

ben-githubs:
### Background

Currently it is difficult to deprecate or remove Panther-managed items from panther analysis, since removing the files from the repo does not remove them from the remote Panther instance. This can result in outdated or broken detections remaining present on customer instances, unless we work directly with customers to help them remove them.

Additionally, customers have long complained about drift between their local repo state and their remote Panther state. Customers expect their repo to be a source of truth, and are often surprised to discover items present in their Panther instance that are not represented in their local repo.

### Changes

This PR introduces a new flag you can pass to the upload command: `--sync`. Passing this flag will trigger PAT to fetch a list of Analysis Item IDs from a customer's remote Panther instance, and compare it to a list of all IDs present in the local repo. Customers can then choose to delete items from their remote which aren't present locally.

The state comparison is performed using Panther's REST API to list data models, rules, helpers, etc. The comparison is shallow - we do not consider the content of the items at all, we just evaluate if they exist in both the remote and the local repo. By default, customers are presented with a list of IDs that are marked for deletion, and prompted whether they would like to continue. This can be bypassed by passing `--no-confirm` to the upload command in addition to `--sync`. The deletion is handled via the same mechanism as PAT's native `delete` command.

If `--sync` is not passed, the default behaviour remains the same - PAT will upload the items in the local repo without deleting anything remotely. This ensures backwards compatibility with any automated workflows customers may have created, which assumed the original behaviour.

### Testing

I've added some unit tests to ensure individual components work as expected, and have done some live testing on a dev Panther instance to ensure remote items are detected and deleted properly.


ben-githubs:
I'm still working on test cases and whatnot, but figured I'd make a draft PR so discussions can start early, if needed.

PR #543, opened 2024-09-16
Add Pagination to Delete Commands

ben-githubs:
### Background

The `delete` command cannot handle more than 100 IDs being passed at any given time, so this PR adds some basic pagination to that command. This is necessary because we intend to start deprecating rules and ask customers to run delete against a masterlist of IDs which will grow over time.

### Changes

* added a batching function
* modified `PublicAPIClient.delete_saved_queries` and `PublicAPIClient.delete_detections` to invoke the GQL endpoint in batches of 100 or less

### Testing

* tested manually by invoking `pat delete` with 150 detection IDs and 150 query names
* added 3 unit tests for the `_batched` function


panther_analysis_tool/backend/public_api_client.py
```diff
@@ -329,23 +329,32 @@ def transpile_filters(
...
         self, params: DeleteSavedQueriesParams
     ) -> BackendResponse[DeleteSavedQueriesResponse]:
-        query = self._requests.delete_saved_queries()
-        delete_params = {
-            "input": {
-                "dryRun": params.dry_run,
-                "includeDetections": params.include_detections,
-                "names": params.names,
+        data: Dict = {"names": [], "detectionIDs": []}
+        # backend's delete function can only handle 100 IDs at a time,
```
le4ker, line 333: This restriction comes from Panther's API, and it's due to the DynamoDB limitation you mention bellow, right?

le4ker (approved):

darwayne (approved):

PR #553, opened 2024-10-11
add option to print only failing unit tests

ben-githubs:
### Background

PAT test currently prints all test results, with an option to sort them to the bottom of the result list. This sucks for github because you have to scroll through thousands of lines in a github action to see what test(s) failed. This PR adds a flag to only print failed test results, so the output is minimal and easy to grep.

### Changes

* introduced `--print-failed-only` flag

### Testing

* `make tests` works
* manually tested with/without flag

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -800,7 +800,9 @@ def test_analysis(
     invalid_specs.extend(invalid_data_models)
 
     all_test_results = (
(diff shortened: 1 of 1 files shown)


arielkr256 (dismissed):
LGTM!

maxrichie5 (approved):
I think I might recommend the name `--show-failures-only` but code looks good

le4ker:
@ben-githubs the flag recommendation from @maxrichie5 is more suitable I believe 😄

arielkr256 (approved):

PR #557, opened 2024-11-18
Fix broken sql table name validation example

dekatzenel:
### Background
https://panther-labs.slack.com/archives/C0719AH18SV/p1731965902688019
<High level overview here>

### Changes

* Adds handling of simple dict table parse case

### Testing

* Added unit test using example provided that failed before and passes now
* Not sure if it works in all cases, but it took a while for any customer to complain about this case, so we can fix future cases as the need arises


maxrichie5 (approved):
I don't really understand how this work but the code seems fine to me

PR #562, opened 2025-01-20
Schema changes should be case-insensitive

papanikge:
## Background

Customers noticed some errors while trying to use `update-custom-schemas` command. 
Turns out it was due to case errors.
Backend already is able to correctly _ignore_ case differences. The error was manifesting due to `revision` conflicts, (because the CLI thought that the schema is new it sent `revision=0`, but the schema existed with `revision > 1+`)

## Changes

* Make local schema checking case-insensitive

[(same as pypanther)](https://github.com/panther-labs/pypanther/pull/196)

### Before
image

### After
image

### with schema being
image

### while at the UI is

image
**and change went thru w/o changing the original name in the db**


nVlast (approved):
Looks good to me. Approved

PR #602, opened 2025-03-26
Make PAT work better on Windows

### LINKED ISSUE #143: Does not install correctly on Windows
### Describe the bug

pip install does not result in an exe file, so will not run

### Steps to reproduce

1. Install via pip or pip3 on Windows 10
2. Attempt to run
3. Get prompt about what to use to open file

### Expected behavior

Run panther_analysis_tool

### Environment

How are you deploying or using Panther?

- Panther version or commit: NA
- OS: Win 10
- Browser: Edge (version NA for this)

### Additional context

Found solution that worked at https://github.com/PyCQA/pydocstyle/issues/73

### Screenshots

image

Screenshot shows the file from install (1 KB unspecified file) and an exe created by renaming one of the other exe files). Renaming the file as an exe did not resolve.

ben-githubs:
### Background

PAT doesn't work well on Windows because you can't natively execute python files; you have to invoke them using the python executable. This is troublesome because you then have to fully specify the path to PAT each time.

Python has a nice feature called `entry_points` which replaces the `scripts` configuration value and is more cross-platform friendly. By using this config parameter instead of scripts, `pip` knows to generate an EXE file for PAT as part of the installation process. You can then add the EXE to your path and call PAT from the command line anywhere in Windows.

Should fix #143!

### Changes

* swapped `scripts` with `entry_points` in setup.py

### Testing

* Tried installing from source in a Windows VM and it worked great!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,12 @@ setup(
     url="https://github.com/panther-labs/panther_analysis_tool",
     download_url=f"https://github.com/panther-labs/panther_analysis_tool/archive/v{PAT_VERSION}.tar.gz",
     keywords=["Security", "CLI"],
-    scripts=["bin/panther_analysis_tool", "bin/pat"],
+    entry_points={
+        "console_scripts": [
+            "panther_analysis_tool=panther_analysis_tool.main:run",
+            "pat=panther_analysis_tool.main:run",
+        ],
+    },
     install_requires=install_requires,
     classifiers=[
         "Development Status :: 4 - Beta",


ben-githubs:
We'll need to update our docs with some windows-specific install instructions, since it still isn't as straightforward as it is in unix/linux. You have to make sure the scripts directory is added to your path and that can be sort of a pain to do if you don't know how

le4ker (approved):

ben-githubs:
This may actually be unnecessary; the Poetry migration PR might do this as well... I'll need to check!

ben-githubs:
Confirmed: the Poetry version does the same thing as this PR. Closing this for now, if we choose not to move forward with Poetry, we can re-open.

PR #636, opened 2025-07-22
Add debug command

ben-githubs:
### Background

Adds a `pat debug` command which allows users to test a single unit test in a debug-friendly environment:

- print statements and other output to stdout are surfaced to user
- breakpoints are respected and users can enter their chosen debugger
- errors are presented with traceback relative to the rule.py file (i..e we don't include the wrapping PAT code in the traceback)
- general test output is suppressed (won't see "PASS" or "FAIL", for example)

Under the hood, I allowed an extra param to be passed to `test_analysis` which is propagated down to the actual testing code. This ensures that test and debug have the same environment.

### Changes

* added new command

### Testing

* manual testing works great
* unit tests incoming

### DIFF
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -14,6 +14,7 @@ import shutil
 import subprocess  # nosec
 import sys
 import time
+import traceback
 import typing  # 'from typing import Optional' conflicts with 'from schema import Optional', below
 import zipfile
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,6 +105,34 @@ $ panther_analysis_tool test --filter RuleID=AWS.CloudTrail.Stopped --test-names
                 [PASS] [rule] false
 ```
 
+### Debug
+
+Run a specific unit test in debug mode:
+
+```bash
+panther_analysis_tool test My.RuleID "My unit test name"
```
arielkr256, line 113: Should this be `pat debug`?

arielkr256: Do we want to allow `pat debug My.RuleID` with no unit tests, to debug all unit tests for a given rule?

ben-githubs: Yeah that should be `debug` instead of `test`; I'll fix that.

I opted not to allow you to debug every test because I think `debug` makes more sense as a targeted command you run on specific input when you're troubleshooting something. I think `test` should be the primary command users invoke to check if their rules work, and `debug` should only be called to investigate why a particular test fails

ben-githubs:
Gonna hold off on merging, because I want to test it with the error-printing changes I just merged. Plus, this would be a minor version update rather than a patch update anyway!

PR #640, opened 2025-07-25
feat: invisible character linting in PR

LucySuddenly:
### Description

Invisible characters represent a risk to public-facing repositories, potentially causing unintended harm or enacting malicious intent. This CI check will eyeball all PRs, on every commit, and submit a failure status which blocks the PR.

Changes to files in the script folder will trigger a basic test, which should not be considered an exhaustive validation of each invisible character, but rather an indication that the core functionality has not regressed.

### References

Inspired by https://github.com/flopp/invisible-characters

### Checklist

- [x] Added unit tests 
- [x] Tested end to end, including ~~screenshots or videos~~ terminal output copy/paste

### Notes for Reviewing

New readme contains instructions for running manually. 

I did a check so we can assume we're clean before this starts running on PRs:

`find . -type f -not -path '*/\.*' -exec python3 .github/scripts/lint-invisible-characters/lint-invisible-characters.py {} +`
`Scanned 300 file(s): no invisible characters found`


arielkr256 (approved):
Cool!

PR #758, opened 2026-02-26
Improve test command performance with faster YAML parsing and parallel sqlfluff validation

tomasz-sq:
## Problem

The `test` command becomes increasingly slow as detection repos grow.

## Root Cause

Profiling revealed that 74% of the runtime was spent in `sqlfluff.parse()` — used to validate table names in scheduled queries. Each call took ~0.2s, all running sequentially inside the classification loop. The remaining time was split between YAML parsing with ruamel.yaml's pure-Python safe loader and schema validation.

## Solution

1. **Faster YAML parsing** — Replaced ruamel.yaml's pure-Python `YAML(typ="safe")` with PyYAML's C-extension `CSafeLoader` for non-roundtrip loading. Added `_PATSafeLoader` subclass to preserve timestamp-as-string behavior.
2. **Parallel sqlfluff validation** — Deferred `contains_invalid_table_names()` out of the classification loop into a batch step. Added `--workers N` flag (`PANTHER_WORKERS` env var) that uses `ProcessPoolExecutor` to run sqlfluff calls in parallel.
3. **O(1) base detection lookup** — Replaced linear scan with a dict index in `setup_run_tests()`.
4. **`analysis_ids` list → set** — O(1) membership checks for duplicate ID detection.

## Results

- Sequential (workers=1): **~13% faster**
- Parallel (workers=4): **~56% faster**

## Test plan
- [x] All 274 unit tests pass (6 new tests added)
- [x] Verified sequential output matches parallel output
- [x] No changes to public API — `--workers` defaults to 1 (existing behavior)

### DIFF
--- a/panther_analysis_tool/analysis_utils.py
+++ b/panther_analysis_tool/analysis_utils.py
@@ -641,7 +641,7 @@ def transpile_inline_filters(
         logging.info("No backend client provided, skipping InlineFilters during testing")
 
 
-def load_analysis(  # pylint: disable=too-many-arguments
+def load_analysis(  # pylint: disable=too-many-arguments,too-many-positional-arguments
     path: str,
     ignore_table_names: bool,
     valid_table_names: List[str],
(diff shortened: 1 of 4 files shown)


zaynahsmith-dasilva (approved):

maxrichie5 (approved):
Very cool additions to make test faster. Thank you for submitting this. I appreciate the comments made and especially the new tests to ensure the parsing works correctly.

PR #764, opened 2026-03-03
Install dependent scheduled queries for scheduled rules

maxrichie5:
### Background

Scheduled rules have scheduled queries they use. If a user installs a scheduled rule, they need to also install the scheduled queries it uses. 

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


darwayne (approved):
Nice, it would probably be good to add a test case for the query not in cache edge case

PR #766, opened 2026-03-05
fix(ci): skip GPG signing when fork PRs need no formatting changes

mbellifa:
### Background

Our formatting step requires a secret to sign commits back to the repo. This causes the check to fail from forks which (desirably) do not have access to our secrets.

### Changes

* Checks for changes from the formatting step before attempting to load gpg secrets
* Explicitly catches and fails the case when a forked repo has formatting changes that cannot be committed back


maxrichie5 (approved):
Makes sense

PR #785, opened 2026-04-08
feat: add AI skills support (AnalysisType: skill) for validation, packaging, and upload

rleighton:
Summary                                                                                                  
   
  Adds a new skill analysis type to the analysis tool so skill YAML definitions can be validated, packaged,
   and bulk-uploaded alongside existing rules, policies, queries, etc. The server-side skills feature lives
   in panther-enterprise (internal/ai/skills/ + internal/core/analysis_api/bulkupload/converter_skills.go);
   this PR is the CLI counterpart.                          

  What's included

  - SKILL_SCHEMA in schemas.py — required: AnalysisType, SkillName (lowercase snake_case), Description,    
  Prompt. Optional: DisplayName, ToolMessage, Enabled, Tags, DependsOn, RequiredTools, Namespace
  (regex-validated). Strict key checking to catch typos.                                                   
  - Core wiring across constants.py, core/definitions.py, analysis_utils.py, util.py, validation.py: new
  AnalysisTypes.SKILL, SKILLS_PATH_PATTERN = "*skills*", skills list on ClassifiedAnalysisContainer,       
  SkillName as the spec ID field, pretty-name support.
  - Packaging (main.py): skills get their own ZipChunk (max 100) in bulk upload, plus a "Skills" row in the
   upload summary.                                                                                         
  - Backend contract (backend/client.py): extends BulkUploadResponse with skills: BulkUploadStatistics.
  - Pack support (detection_schemas/analysis_config_schema.json): "skill" added to the pack AnalysisType   
  enum.                                                                                                    
  - Tests: new tests/fixtures/skills/example_skill.yml fixture; 14 schema test cases; zip-chunk count      
  bumped 7→8; mocked bulk-upload responses extended with skills=stats.                                     
                                                            
  Notes on server alignment                                                                                
                                                            
  Schema matches the current server SkillSpec (internal/ai/skills/yml/skill.go) as of the latest sync:     
  - RequiredTools is included.                                                                             
  - Namespace uses the same ^[a-z][a-z0-9_]*$ pattern as toolnames.NamespacePattern.
                                                                                                           
  Byte-length caps and DependsOn count/normalization rules are intentionally left to server-side validation
   — the bulk-upload converter enforces them via limits.ValidateFieldSizes and                             
  toolnames.NormalizeAndValidateDeps, and the analysis tool doesn't attempt to mirror them to avoid drift.
                                                                                                           
  DependsOn existence is not checked locally; the server's bulk-upload path also skips eager               
  existence/cycle checks by design (see converter_skills.go:73-87) so that authors can import a folder of
  mutually-dependent skills in one batch regardless of ZIP order.


rleighton:
@cursor review

rleighton:
@cursor review

rleighton:
@cursor review

rleighton:
@cursor review

rleighton:
@cursor review

rleighton:
@claude review

panther_analysis_tool/schemas.py
```diff
@@ -328,6 +329,24 @@ def validate(  # type: ignore
...
+        "SkillName": And(str, Regex(r"^[a-z][a-z0-9_]*$")),
+        "Description": str,
+        "Prompt": str,
+        Optional("DisplayName"): str,
+        Optional("ToolMessage"): str,
+        Optional("Enabled"): bool,
+        Optional("EagerLoading"): bool,
+        Optional("Tags"): [str],
+        Optional("DependsOn"): [str],
+        Optional("RequiredTools"): [str],
```
arielkr256: **Finding: tool names are not validated anywhere — client or server.**

Tested against the live threat-research instance: all of the following passed both `validate` and a real `upload`:
- nonexistent tools in `RequiredTools` (`totally_fake_tool_that_does_not_exist`)
- nonexistent tools referenced as `{tool:...}` in `Prompt`
- glob/regex patterns: `panther_ai_*`, bare `*`, `.*`, `[a-z]+_tool`, `panther_ai_(alerts|detections)_list`

So a typo like `panther_ai_alert_list` (singular) sails through CI and upload, and only surfaces at skill runtime.

Suggestions:
- Cheap client-side guard in this PR: apply the same `^[a-z][a-z0-9_]*$` regex used for `SkillName` to each `RequiredTools` entry — rejects wildcards, regex syntax, and casing mistakes. (First confirm with backend whether wildcards are intentionally supported; if so, allow `*` in the pattern.)
- Optionally lint that every `{tool:...}` name in `Prompt` appears in `RequiredTools` or matches the format regex.
- Real existence checking against the tool registry has to live server-side (backend ticket) — only the server knows the valid tool list.

panther_analysis_tool/backend/graphql/bulk_upload.graphql
```diff
@@ -21,6 +21,9 @@ mutation UploadDetections($input: UploadDetectionEntitiesInput!) {
         correlationRules {
             ...UploadStatisticDetails
         }
+        skills {
```
arielkr256, line 24: Verified live against threat-research: these stats come back correctly — uploading panther-analysis#2091 reported `Skills: Total 13, New 11, Modified 2`, matching the 11 new skills plus 2 pre-existing Okta skills exactly.

One server-side observation while testing: re-uploading byte-identical skill files still reports them as `Modified` (unchanged data models/helpers correctly report `Modified: 0`). Not a bug in this PR — PAT just reports what the server returns — but worth flagging to backend if no-op detection is intended for skills.

panther_analysis_tool/schemas.py
```diff
@@ -328,6 +329,24 @@ def validate(  # type: ignore
     ignore_extra_keys=False,
 )  # Prevent user typos on optional fields
 
+SKILL_SCHEMA = Schema(
```
arielkr256, line 384: Field coverage sanity check: panther-analysis PRs #2091/#2098/#2072 collectively exercise 11 of the 12 fields here (everything except `Namespace`), and with `ignore_extra_keys=False` their clean validation passes confirm the schema matches what skill authors are actually writing. Also verified the `SkillName` regex rejects bad input (`BadName-Upper` correctly flagged).

arielkr256 (commented):
## End-to-end test report

Tested this branch against three real panther-analysis PRs that add skills — [#2091 (PantherAI Skills, 11 files)](https://github.com/panther-labs/panther-analysis/pull/2091), [#2098 (GitHub security audit skill)](https://github.com/panther-labs/panther-analysis/pull/2098), and [#2072 (Okta skills, merged)](https://github.com/panther-labs/panther-analysis/pull/2072) — both offline and live against the threat-research instance.

### Offline (`test`, `zip`, `fmt`)
- `test --path skills`: 0 invalid on all three PRs. Confirmed validation genuinely runs (not silently skipped) by corrupting a `SkillName` to `BadName-Upper` — correctly rejected against `^[a-z][a-z0-9_]*$`.
- Schema field coverage: the three PRs collectively use 11 distinct top-level fields (`AnalysisType`, `SkillName`, `Description`, `Prompt`, `DisplayName`, `ToolMessage`, `Enabled`, `EagerLoading`, `Tags`, `DependsOn`, `RequiredTools`). Since `SKILL_SCHEMA` sets `ignore_extra_keys=False`, clean passes mean every field is recognized. Only `Namespace` went unexercised by real-world data.
- `zip`: all skill files correctly included in the archive on every PR (e.g. 13/13 for #2091's branch).
- `fmt`: no unwanted modifications to skill YAML.
- Full-repo packaging on #2091: 902 passed / 0 failed; the only invalids were a missing `panther_detection_helpers` pip dep, unrelated to skills.

### Live (`validate`, `upload` against threat-research)
- Permission enforcement: with a token lacking `ManageAISkills`, `validate` failed with per-file errors correctly attributed to each skill filename (`ManageAISkills permission required to upload skills`), while non-skill paths validated fine. Good error-mapping behavior.
- After granting the permission: `validate` and `upload` succeeded on all three PRs, and the new skills upload stats came back exactly right — #2091: `Total 13, New 11, Modified 2`; #2098: `Total 3, New 1, Modified 2`; #2072: `Total 2, New 0, Modified 2`.

### Findings / follow-ups (none blocking)
1. **No tool-name validation anywhere** — see inline comment on `RequiredTools` in `schemas.py`.
2. **Server counts unchanged skills as Modified** — re-uploading byte-identical skill files back-to-back still reports `Modified: 2`, while unchanged data models/helpers correctly report `Modified: 0`. Server-side behavior, not this PR's bug, but worth a backend ticket if no-op detection is intended for skills.
3. **`delete` has no skills support** (only `--analysis-id`/`--query-id`). Fine if intentionally out of scope, but there's currently no PAT way to remove an uploaded skill.
4. Pre-existing cosmetic bug (not from this branch): `test` prints `Passed: -1` when a directory contains invalid specs but no runnable tests — `main.py:1291` subtracts invalid specs from the test count.

panther_analysis_tool/schemas.py
itsmostafa: Added a client-side guard in fd0576b: each RequiredTools entry now must match the SkillName regex `^[a-z][a-z0-9_]*$`, so typos, casing mistakes, wildcards, and regex patterns are rejected at validate/upload. Confirmed with backend that wildcards are not supported, so the strict pattern is correct. Real existence checking against the tool registry stays server-side.

panther_analysis_tool/backend/graphql/bulk_upload.graphql
itsmostafa: So skills report `Modified` on a no op re upload while other entity types correctly report `Modified`: 0. Is that a known gap in skills' diffing logic, or expected because skill metadata (like timestamps) always changes server side?

panther_analysis_tool/schemas.py
arielkr256: If wildcards are not supported, what is the proper way to reference toolnames for BYO MCPs, which have customer-defined names?

arielkr256: > Real existence checking against the tool registry stays server-side.

Has this been implemented server-side?

itsmostafa: yes, it is implemented server-side, but at skill-load/inference time rather than at upload. `ToolBox.hasRequiredTools` compiles each pattern and requires it to match at least one registered tool; skills with unmet required tools are not loaded for inference. Upload itself does not reject a pattern that matches no tool, so a valid but nonexistent pattern uploads fine and simply results in the skill not loading. I've updated the client guard to mirror this: each entry is validated as a compilable regex rather than forced to snake_case, so only non-compilable or empty entries are rejected.

panther_analysis_tool/schemas.py
```diff
@@ -357,6 +379,24 @@ def validate(  # type: ignore
...
+        "SkillName": And(str, Regex(r"^[a-z][a-z0-9_]*$")),
+        "Description": str,
+        "Prompt": str,
+        Optional("DisplayName"): str,
+        Optional("ToolMessage"): str,
+        Optional("Enabled"): bool,
+        Optional("EagerLoading"): bool,
+        Optional("Tags"): [str],
+        Optional("DependsOn"): [str],
+        Optional("RequiredTools"): [And(str, validate_required_tool_pattern)],
```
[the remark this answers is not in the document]
itsmostafa: This is by design. RequiredTools entries are regex patterns, not literal tool names, matched server-side by toolnames.CompileRequiredToolPattern which compiles each as ^<pattern>$. This supports BYO MCPs with customer-defined names. The client guard mirrors the server: it rejects only entries that fail to compile as a regex (or are empty), and real existence checking stays server-side at skill-load time via ToolBox.hasRequiredTools. Rejecting wildcards here would break valid patterns the server accepts.

arielkr256 (approved):

PR #808, opened 2026-06-12
feat: add validate-sql command to compile query SQL against the data lake

### LINKED ISSUE #639: Add SQL Validation for Scheduled Queries Using Panther Backend
### Describe the ideal solution

Add SQL validation for scheduled queries that executes the SQL against Panther's backend data lake to catch syntax errors, invalid table references, and other runtime issues that static analysis tools like `sqlfluff` might miss. This would leverage existing GraphQL mutations (`executeDataLakeQuery` and `getDataLakeQuery`) to validate SQL syntax by executing queries with minimal resource consumption (e.g., `LIMIT 1`).

### Describe your use cases

1. **Pre-deployment validation**: Validate scheduled query SQL before deploying to production to catch errors early
2. **CI/CD integration**: Automate SQL validation in deployment pipelines to prevent broken queries from reaching production
3. **Development workflow**: Provide immediate feedback during development when writing or modifying scheduled queries
4. **Schema validation**: Ensure table and column references are valid against the actual Panther data lake schema
5. **SQL syntax validation**: Catch SQL syntax errors that static analysis tools might not detect

### How frequently would you use such a feature

Daily - This would be used during development of new scheduled/changes to existing queries and as part of the CI/CD pipeline for every deployment.

### Describe alternatives you have considered

1. **Manual testing**: Currently requires manual execution in Panther UI, which is time-consuming and error-prone
2. **Custom validation scripts**: Would require maintaining separate validation infrastructure
3. **Backend API validation**: Using existing Panther GraphQL mutations provides a more integrated solution

### References

N/A

### Additional context

The Panther MCP repository already contains the necessary GraphQL mutations (`executeDataLakeQuery` and `getDataLakeQuery`) that could be leveraged for this validation. The implementation would involve:
1. Extracting SQL from scheduled query YAML files
2. Executing SQL with LIMIT 1 via executeDataLakeQuery mutation
3. Checking query status via getDataLakeQuery to determine if SQL is valid
4. Reporting validation results back to the user

This approach would provide more comprehensive validation than static analysis alone while leveraging existing Panther infrastructure.

arielkr256:
## Summary

Adds a `validate-sql` command that validates the SQL of scheduled queries, saved queries, and SQL-based lookup tables by compiling them against the Panther data lake. Each query is executed with an `EXPLAIN` prefix via the `executeDataLakeQuery` / `dataLakeQuery` public API operations — Snowflake compiles the query without scanning any data (0 bytes scanned), surfacing syntax errors and invalid table/column references that static analysis (sqlfluff) misses.

Closes #639

```
panther_analysis_tool validate-sql --path . --api-token ... --api-host ...
```

## Behavior

- **Validated**: `scheduled_query` / `saved_query` specs (`Query` or `SnowflakeQuery`) and `lookup_table` specs with a `Query` field. Supports the standard `--filter` / `--ignore-files` options.
- **Missing tables** (log source not onboarded in the instance):
  - Enabled queries **fail** — a scheduled query that can't resolve its table is broken in that instance.
  - Disabled queries are **skipped** (syntax errors still fail).
  - `--skip-missing-tables` extends the skip to enabled queries, for repos that ship content for log sources the validating instance doesn't have.
- **Skipped**: Jinja macro library files (`{% macro %}`-only SQL, which expands to nothing standalone; queries that *use* macros validate fine since the backend expands templates server-side) and investigation templates with `<PLACEHOLDER>` markers.

## Live testing

Ran against panther-analysis (106 queries) on the threat-research instance (~3.5 min):

- Found a **real bug**: `kubernetes_ioc_activity_query.yml` has its `INNER JOIN` after the `WHERE` clause — invalid SQL never caught by static analysis (filed as panther-labs/panther-analysis#2101).
- Confirmed `p_occurs_since` and Jinja macros expand server-side under `EXPLAIN`.
- All skip/fail categories verified against real instance responses.

## Implementation notes

- New GraphQL operations follow the existing pattern: `.graphql` files + `PublicAPIRequests` loaders + typed params/response dataclasses + `Client` abstract methods. `LambdaClient` raises (unsupported), consistent with other public-API-only features.
- Validation polls `dataLakeQuery` until a terminal status with a 60s per-query timeout.
- Also excludes `venv` from black/isort so `make fmt` doesn't reformat an in-project venv (separate commit).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/panther_analysis_tool/command/validate_sql.py
+++ b/panther_analysis_tool/command/validate_sql.py
@@ -27,6 +27,10 @@ _MISSING_TABLE_ERROR = "does not exist or not authorized"
 # template queries, which are meant to be filled in by a human before running.
 _PLACEHOLDER_PATTERN = re.compile(r"<[A-Z][A-Z0-9_]*>")
 
+# The backend routes queries with this pragma to the PantherFlow engine, which
+# ignores the EXPLAIN prefix and executes the query for real.
+_PANTHERFLOW_PRAGMA = re.compile(r"--\s*pragma:\s*pantherflow", re.IGNORECASE)
+
 
 @dataclass
 class ValidateSqlArgs:
@@ -122,6 +126,13 @@ def _collect_targets(args: ValidateSqlArgs) -> List[SqlValidationTarget]:
         # nothing on their own and only compile when included from another query.
         if "{% macro" in sql:
             logging.info("Skipping %s: query only defines jinja macros", analysis_id)
+            continue
+
+        if _PANTHERFLOW_PRAGMA.search(sql):
+            logging.info(
+                "Skipping %s: PantherFlow queries cannot be compile-checked with EXPLAIN",
+                analysis_id,
+            )
             continue
 
         placeholder = _PLACEHOLDER_PATTERN.search(sql)
--- a/panther_analysis_tool/command/validate_sql.py
+++ b/panther_analysis_tool/command/validate_sql.py
@@ -1,0 +1,187 @@
+import logging
+import re
+import time
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Tuple
+
+from panther_analysis_tool import cli_output
+from panther_analysis_tool.analysis_utils import filter_analysis, load_analysis_specs
+from panther_analysis_tool.backend.client import Client as BackendClient
+from panther_analysis_tool.backend.client import (
+    DataLakeQueryStatus,
+    ExecuteDataLakeQueryParams,
+    GetDataLakeQueryParams,
+    UnsupportedEndpointError,
+)
+from panther_analysis_tool.constants import AnalysisTypes
+from panther_analysis_tool.core.definitions import ClassifiedAnalysis
+from panther_analysis_tool.core.parse import Filter
+
+POLL_INTERVAL_SECONDS = 1
+PER_QUERY_TIMEOUT_SECONDS = 60
+
+_SQL_KEYS = ("Query", "SnowflakeQuery")
+_MISSING_TABLE_ERROR = "does not exist or not authorized"
+
+# Matches substitution markers like <ACTOR_ID> or <WINDOW> used by investigation
+# template queries, which are meant to be filled in by a human before running.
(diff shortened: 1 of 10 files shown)


panther_analysis_tool/command/validate_sql.py
```diff
@@ -0,0 +1,198 @@
...
+    for target in targets:
+        try:
+            error = _validate_target(backend, target)
+        except UnsupportedEndpointError as err:
+            logging.debug(err)
+            return 1, cli_output.warning("Your Panther instance does not support this feature")
+
+        if error is None:
+            print(f"  {cli_output.success('PASS')} {target.analysis_id}")
+        elif _MISSING_TABLE_ERROR in error and (args.skip_missing_tables or not target.enabled):
```
dekatzenel, line 71: The missing-table heuristic over-matches and silently hides real bugs.
_MISSING_TABLE_ERROR = "does not exist or not authorized" is a substring match, but Snowflake uses that exact phrasing for functions, views, stages, schemas — not just tables (e.g. Function 'FOO' does not exist or not authorized, Object 'DB.SCHEMA' does not exist or not authorized). So a query with a typo'd UDF name, a dropped view, or a wrong schema gets classified as a "missing table" and SKIPPED whenever the query is disabled (default) or --skip-missing-tables is set — instead of FAILing.

panther_analysis_tool/command/validate_sql.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if _PANTHERFLOW_PRAGMA.search(sql):
+            logging.info(
+                "Skipping %s: PantherFlow queries cannot be compile-checked with EXPLAIN",
+                analysis_id,
+            )
+            continue
+
+        placeholder = _PLACEHOLDER_PATTERN.search(sql)
+        if placeholder:
```
dekatzenel, line 139: Should these be counted in the summary?

dekatzenel (commented):
Cursor comments look worth fixing, plus a couple of other notes. Can you add screenshots of the success and failure output to the top-level comment?

PR #813, opened 2026-06-25
feat: add scheduled prompt support (AnalysisType: scheduled_prompt) for validation, packaging, and upload

kdmnt:
## What this adds

Panther AI **scheduled prompts** run on a schedule as a chosen user or API token — today they can only be authored in the Console. This PR teaches the analysis tool to treat a `scheduled_prompt` YAML like any other analysis item, so prompts can be **validated, packaged, and bulk-uploaded through CI/CD** the same way rules and queries already are.

The server side ships in three PRs (links are informational — `panther-enterprise` is private; this PR is self-contained and reviewable from the diff alone):

- **1/3** — https://github.com/panther-labs/panther-enterprise/pull/29629: the `ViewScheduledPrompts` / `ManageScheduledPrompts` permissions.
- **2/3** — https://github.com/panther-labs/panther-enterprise/pull/29634: the YAML model, store, and upsert client.
- **3/3** — https://github.com/panther-labs/panther-enterprise/pull/29754: the bulk upload/download pipeline and Console UI — the endpoints this PR calls.

## YAML shape

```yaml
AnalysisType: scheduled_prompt
PromptName: weekly_iam_review          # unique id; must start with a lowercase letter, then lowercase letters / digits / underscores
DisplayName: Weekly IAM Review
PromptText: Summarize new IAM users created this week.
RunAsUser: alice@example.com           # required; a user email OR a Panther API-token id (begins with po_)
Schedule:                              # exactly one of CronExpression / RateMinutes
  CronExpression: "0 9 * * 1"
  TimeoutMinutes: 10
# optional: Description, OutputLength (small|medium|largest), Enabled, Private (false only — true is rejected)
```

## Changes

- **`schemas.py`** — `SCHEDULED_PROMPT_SCHEMA`, with strict key checking (typos are rejected, not silently ignored).
- **Core wiring** (`constants.py`, `core/definitions.py`, `analysis_utils.py`, `util.py`) — the new analysis type, a `*scheduled_prompts*` path glob for discovery, `PromptName` as the identifier, and classification + display-name support.
- **`validation.py`** — `validate_packs` now derives each item's id via `item.analysis_id()` instead of a hardcoded field chain that ended in a bare `LookupName` subscript — which would `KeyError` on any type without that field (including `scheduled_prompt`).
- **`main.py`** — prompts get their own upload chunk and a `Scheduled Prompts` row in the upload summary.
- **`backend/client.py` + the two bulk-upload `.graphql` files** — the `.graphql` files add `scheduledPrompts` to the upload mutation and status query; `client.py` parses the counts. Without the field the server omits it and the counts silently read as zero, so a dedicated test (`test_graphql_queries.py`) guards its presence.
- **`analysis_config_schema.json`** — adds `scheduled_prompt` to the `AnalysisType` enum, so it's recognized in config-file validation (used by `pat test` and pack validation).
- Fixtures and unit tests.

## Validation contract

- **`RunAsUser` is required and polymorphic** — a user email or a `po_` token id. The client only checks it is non-empty; existence, liveness, and the `RunPantherAI` permission are validated server-side and surfaced by `pat validate`.
- **`Private` is shared-only** — `true` is rejected (not silently ignored); prompts are made private only via the Console.
- Numeric bounds are server-authoritative — the `RateMinutes ≥ 5` floor, the `TimeoutMinutes` 1–15 range, and text-length limits are enforced server-side and surfaced by `pat validate`. (The client applies only the generic schedule checks it shares with saved queries — `RateMinutes > 1`, `RateMinutes ≥ TimeoutMinutes` — so e.g. `RateMinutes: 3` passes locally but is rejected on upload.)

## Not in scope: deletion

There is no `pat delete` for scheduled prompts — matching **AI Skills** (#785), which also ships upload/validate-only. Both lack a public delete mutation today; a delete path can be added for prompts and skills together once one exists. To take a prompt out of rotation as-code, set `Enabled: false` and re-upload; hard-delete is via the Console.

## Auth: token only

Uploading scheduled prompts requires **API-token auth** (`--api-token` + `--api-host`). **Legacy AWS-profile auth** (no token; direct `panther-analysis-api` Lambda invoke via `--aws-profile` / the boto3 default chain) is rejected per file with `system user cannot import scheduled prompts` — that path carries no user/token actor, and a prompt must have a real owner. Detections still upload under legacy auth; prompts are the exception. (`pat validate` is token-only regardless.)

## Testing

Unit tests cover the schema contract (required/optional fields, `RunAsUser` polymorphism, `Private` rejection), packaging, classification, the upload summary, the GraphQL-field guard, and a `validate_packs` regression (a prompt next to a pack must not crash ID resolution). Verified live against a dev tenant: `pat validate` + `pat upload` + idempotent re-import, with run-as accepted for a permitted user/token and rejected for one lacking `RunPantherAI` or no longer active.

```bash
make install && make test
# local, no instance (schema-validates the fixture; prompts aren't executable, so the test runner doesn't run them):
pat test     --path tests/fixtures/scheduled_prompts
# validate — no writes; token needs BulkUploadValidate (or BulkUpload) + ManageScheduledPrompts:
pat validate --path tests/fixtures/scheduled_prompts --api-host "$HOST" --api-token "$TOK"
# upload — token needs BulkUpload + ManageScheduledPrompts (+ AIRunAsModify only when RunAsUser is a different identity than the token):
pat upload   --path tests/fixtures/scheduled_prompts --api-host "$HOST" --api-token "$TOK"
```

### DIFF
--- a/panther_analysis_tool/constants.py
+++ b/panther_analysis_tool/constants.py
@@ -15,6 +15,7 @@ from panther_analysis_tool.schemas import (
     POLICY_SCHEMA,
     RULE_SCHEMA,
     SAVED_QUERY_SCHEMA,
+    SCHEDULED_PROMPT_SCHEMA,
     SCHEDULED_QUERY_SCHEMA,
 )
 
@@ -32,6 +33,7 @@ PACKS_PATH_PATTERN = "*/packs"
 POLICIES_PATH_PATTERN = "*policies*"
 QUERIES_PATH_PATTERN = "*queries*"
 RULES_PATH_PATTERN = "*rules*"
+SCHEDULED_PROMPTS_PATH_PATTERN = "*scheduled_prompts*"
 
 
 class AnalysisTypes:
@@ -47,6 +49,7 @@ class AnalysisTypes:
     SCHEDULED_RULE = "scheduled_rule"
     SIMPLE_DETECTION = "simple_detection"
     CORRELATION_RULE = "correlation_rule"
+    SCHEDULED_PROMPT = "scheduled_prompt"
 
 
 # The UserID is required by Panther for some API calls, but we have no way of
@@ -66,6 +69,7 @@ SCHEMAS: Dict[str, Schema] = {
     AnalysisTypes.DERIVED: DERIVED_SCHEMA,
     AnalysisTypes.SCHEDULED_RULE: RULE_SCHEMA,
     AnalysisTypes.CORRELATION_RULE: CORRELATION_RULE_SCHEMA,
+    AnalysisTypes.SCHEDULED_PROMPT: SCHEDULED_PROMPT_SCHEMA,
 }
 
 SET_FIELDS = [
--- a/panther_analysis_tool/main.py
+++ b/panther_analysis_tool/main.py
@@ -283,6 +283,7 @@ def zip_analysis_chunks(
         ZipChunk(patterns=[], types=AnalysisTypes.SCHEDULED_RULE, max_size=200),  # type: ignore
         ZipChunk(patterns=[], types=AnalysisTypes.LOOKUP_TABLE, max_size=100),  # type: ignore
         ZipChunk(patterns=[], types=AnalysisTypes.CORRELATION_RULE, max_size=200),  # type: ignore
+        ZipChunk(patterns=[], types=AnalysisTypes.SCHEDULED_PROMPT, max_size=100),  # type: ignore
     ]
 
     filenames = []
@@ -469,6 +470,7 @@ def print_upload_summary(response: dict) -> None:
         "Global Helpers": response.get("global_helpers", {}),
         "Lookup Tables": response.get("lookup_tables", {}),
         "Correlation Rules": response.get("correlation_rules", {}),
+        "Scheduled Prompts": response.get("scheduled_prompts", {}),
     }
 
     for category, stats in categories.items():
--- a/panther_analysis_tool/analysis_utils.py
+++ b/panther_analysis_tool/analysis_utils.py
@@ -39,6 +39,7 @@ from panther_analysis_tool.constants import (
     POLICIES_PATH_PATTERN,
     QUERIES_PATH_PATTERN,
     RULES_PATH_PATTERN,
+    SCHEDULED_PROMPTS_PATH_PATTERN,
     SCHEMAS,
     VERSION_STRING,
     AnalysisTypes,
@@ -423,6 +424,7 @@ def load_analysis_specs_ex(
                             PACKS_PATH_PATTERN,
                             POLICIES_PATH_PATTERN,
                             QUERIES_PATH_PATTERN,
+                            SCHEDULED_PROMPTS_PATH_PATTERN,
                         )
                     )
                 ):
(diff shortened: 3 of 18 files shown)


panther_analysis_tool/constants.py
```diff
@@ -32,6 +33,7 @@
 POLICIES_PATH_PATTERN = "*policies*"
 QUERIES_PATH_PATTERN = "*queries*"
 RULES_PATH_PATTERN = "*rules*"
+SCHEDULED_PROMPTS_PATH_PATTERN = "*scheduled_prompts*"
```
dekatzenel, line 36: It looks like this isn't in panther-analysis yet. Are we sure this is right?

panther_analysis_tool/main.py
```diff
@@ -283,6 +283,7 @@ def zip_analysis_chunks(
         ZipChunk(patterns=[], types=AnalysisTypes.SCHEDULED_RULE, max_size=200),  # type: ignore
         ZipChunk(patterns=[], types=AnalysisTypes.LOOKUP_TABLE, max_size=100),  # type: ignore
         ZipChunk(patterns=[], types=AnalysisTypes.CORRELATION_RULE, max_size=200),  # type: ignore
+        ZipChunk(patterns=[], types=AnalysisTypes.SCHEDULED_PROMPT, max_size=100),  # type: ignore
```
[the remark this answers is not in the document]
dekatzenel: Substrings are ordered, so I think this is wrong, but it might be a real issue for scheduled and correlation rules. Out of scope for this PR.

dekatzenel (commented):
Can you test this using legacy auth? I'm curious what would happen and we might need to document if it doesn't work. To use legacy auth:

1. skip the `validate` command - it's not supported for legacy, only for token
2. Don't provide an api token or host
3. Pick the AWS profile with --aws-profile <name> (or env var PANTHER_AWS_PROFILE). If you omit that too, boto3's default credential chain is used (env vars → default profile → instance role). The profile must point at the AWS account where Panther is deployed and be able to invoke the panther-analysis-api Lambda.

You can confirm that you're using legacy auth if the `INFO: Panther Public API endpoint: $URL` log line is omitted from pat upload

panther_analysis_tool/constants.py
kdmnt: I am really not sure. What should be (or not?) in panther-analysis? 😅

dekatzenel: @panther-labs/threat-research owns that repo and would be the folks to ask

kdmnt: Would this be for "Managed" Scheduled Prompts perhaps? If that's the case, there isn't any of that at the moment.
Does this block this PR though?

kdmnt:
> Can you test this using legacy auth? I'm curious what would happen and we might need to document if it doesn't work. To use legacy auth:
> 
> 1. skip the `validate` command - it's not supported for legacy, only for token
> 2. Don't provide an api token or host
> 3. Pick the AWS profile with --aws-profile  (or env var PANTHER_AWS_PROFILE). If you omit that too, boto3's default credential chain is used (env vars → default profile → instance role). The profile must point at the AWS account where Panther is deployed and be able to invoke the panther-analysis-api Lambda.
> 
> You can confirm that you're using legacy auth if the `INFO: Panther Public API endpoint: $URL` log line is omitted from pat upload

Confirmed, it doesn't work:
image

panther_analysis_tool/constants.py
dekatzenel: It would be best to confirm this value with threat research before merging, just to make sure they won't want to use a different value later

dekatzenel (approved):
Double check the path name with TR and other than that you should be good to go

panther_analysis_tool/constants.py
kdmnt: merging now, and letting threat-research know. If it needs to be changed, we ll do a follow up

arielkr256: this looks reasonable to me! thanks for checking in

PR #833, opened 2026-09-03
build and publish a wheel

dimbleby:
Currently you are only publishing an sdist.  Publish a wheel also.


dimbleby:
I will not be signing a CLA.  You are welcome to this contribution, take it or leave it.

